PR #9, opened 2017-10-25
Feature/settgast/install all headers

rrsettgast:
The CHAI installation did not contain the CHAI headers. 

`install(DIRECTORY ${PROJECT_BINARY_DIR}/include DESTINATION . OPTIONAL)`

was added to copy the entire include directory from the build directory to the install directory

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -97,9 +97,7 @@ install(
   FILES ${PROJECT_BINARY_DIR}/share/chai/cmake/chai-config.cmake
   DESTINATION share/chai/cmake/)
 
-install(
-  FILES ${PROJECT_BINARY_DIR}/include/chai/config.hpp
-  DESTINATION include/chai)
+install(DIRECTORY ${PROJECT_BINARY_DIR}/include DESTINATION . OPTIONAL)
 
 install(
   TARGETS chai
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -110,6 +108,7 @@ install(
 
 install(EXPORT chai-targets DESTINATION share/chai/cmake/)
 
+
```
davidbeckingsale: What's this for?

src/CMakeLists.txt
```diff
@@ -97,9 +97,7 @@ install(
   FILES ${PROJECT_BINARY_DIR}/share/chai/cmake/chai-config.cmake
   DESTINATION share/chai/cmake/)
 
-install(
-  FILES ${PROJECT_BINARY_DIR}/include/chai/config.hpp
-  DESTINATION include/chai)
+install(DIRECTORY ${PROJECT_BINARY_DIR}/include DESTINATION . OPTIONAL)
```
davidbeckingsale: I don't think we want this to be optional.

rrsettgast: True. I’ll fix this.

src/CMakeLists.txt
rrsettgast: Errr.  Uhhhh. Just decided to insert a random line break??  I’ll remove it.

davidbeckingsale:
Fixed by #11

Issue #16, opened 2017-12-21
Need testing on ManagedArray<const T>


DavidPoliakoff:
```
#include <chai/ManagedArray.hpp>
int main(){
  chai::ManagedArray<const double> foo(50);
  const double* foo2 = (const double*) foo;
}
```

This will cause an error in the casting operator, as registerTouch is ill-defined:

```
ArrayManager.hpp:113:8: note: candidate function not viable: no known conversion from 'const double *' to 'void *' for 1st argument;
```

I'm writing workarounds, but it'd be nice for this to work.

DavidPoliakoff:
Note: the fix turned out to be trivial, feel free to merge the PR or use it as a template

davidbeckingsale:
Closed by #17

PR #17, opened 2017-12-21
Feature/const t cast

DavidPoliakoff:
Addresses #16

Don't know about the wisdom of casting to a const double*, but it is now technically feasible

### DIFF
--- a/src/ManagedArray.inl
+++ b/src/ManagedArray.inl
@@ -63,7 +63,7 @@ CHAI_INLINE
 CHAI_HOST_DEVICE ManagedArray<T>::ManagedArray(
     uint elems, ExecutionSpace space):
   m_active_pointer(nullptr),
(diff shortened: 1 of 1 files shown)


src/ManagedArray.inl
```diff
@@ -63,7 +63,7 @@ CHAI_INLINE
 CHAI_HOST_DEVICE ManagedArray<T>::ManagedArray(
     uint elems, ExecutionSpace space):
   m_active_pointer(nullptr),
-  m_resource_manager(ArrayManager::getInstance()),
```
davidbeckingsale: I've been bitten by this before - this was how I ended up with the map find() call failing....

DavidPoliakoff: yeah, I actually think I ran into something analogous in my development, but had forgotten about it

PR #23, opened 2018-02-16
Proposing ManagedMultiArray class

rrsettgast:
I have something like this in my code that provides an interface for working with multidimensional arrays. I thought it might be useful for others, so I extracted the allocation portions of the wrapper. The only additional feature of this over ManagedArray is that it carries the dimensional information, and it provides allocate and reallocate functions with a multidimensional interface. It could be used to seed a RAJA::View/Layout standalone as it has the dim data. Let me know if you think that this is worth adding into CHAI.


davidbeckingsale:
Declining this after our phone conversation.

PR #24, opened 2018-03-27
Efficient realloc: only copy data in touched spaces

davidbeckingsale:


davidbeckingsale:
Thanks @robinson96, I'll write up some more test cases and check that this all works.

src/ArrayManager.inl
```diff
@@ -137,7 +137,9 @@ void* ArrayManager::reallocate(void* pointer, size_t elems)
       pointer_record->m_user_callback(ACTION_ALLOC, ExecutionSpace(space), sizeof(T) * elems);
       void* new_ptr = m_allocators[space]->allocate(sizeof(T)*elems);
 
-      rm.copy(old_ptr, new_ptr);
+      if (pointer_record->m_touched[space]) {
+        rm.copy(old_ptr, new_ptr);
```
robinson96: I assume the rm knows the lengths and this does a reasonable thing here?

davidbeckingsale: Yes sir.

robinson96:
One other concern - by choosing not to do a realloc on the CPU side of things, you're forcing a copy to always happen, while most realloc implementations will avoid a copy if the realloc ends up with the same address (which happens pretty often, in my experience).  Does Umpire have a realloc API? Should it?

davidbeckingsale:
It does, but that means we can't avoid the unnecessary copies on the CPU side - you call realloc and the semantics of realloc are what you will get. If you only want to avoid a device-side copy then that would be fine.

PR #30, opened 2018-05-23
Bugfix/kunen fixes

ajkunen:
Fixed a bunch of warnings with CUDA

Fixed a warning with the implicit conversion to ManagedArray<const T> from ManagedArray<T>.

Fixed a memory error in the ArrayManager reallocate function.

### DIFF
--- a/src/ArrayManager.inl
+++ b/src/ArrayManager.inl
@@ -127,6 +127,9 @@ void* ArrayManager::reallocate(void* pointer, size_t elems)
       return pointer_record->m_pointers[my_space];
     }
   }
+  
+  // only copy however many bytes overlap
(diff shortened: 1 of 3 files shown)


src/ArrayManager.inl
```diff
@@ -127,6 +127,9 @@ void* ArrayManager::reallocate(void* pointer, size_t elems)
       return pointer_record->m_pointers[my_space];
     }
   }
+  
+  // only copy however many bytes overlap
+  size_t num_bytes_to_copy = std::min(sizeof(T)*elems, pointer_record->m_size);
```
davidbeckingsale: 👍

PR #34, opened 2018-06-07
added compile time option to perform range checking on ManagedArray

rrsettgast:


rrsettgast:
Hey...what happened to the travis tests?

src/ManagedArray.inl
```diff
@@ -207,6 +207,9 @@ template<typename T>
 template<typename Idx>
 CHAI_INLINE
 CHAI_HOST_DEVICE T& ManagedArray<T>::operator[](const Idx i) const {
+#if defined(CHAI_ARRAY_BOUNDS_CHECK)
+  assert( i>=0 && static_cast<size_t>(i) < m_elems );
```
davidbeckingsale: Can you make assert work on the GPU? I have a branch with this functionality, but couldn't get it working.

davidbeckingsale:
https://github.com/LLNL/CHAI/tree/feature/CHAI-44-debug-range-check

src/ManagedArray.inl
rrsettgast: Sorry. I didn't see that. Where is HOST_DEVICE_ASSERT defined? I can't find it.

I actually hadn't tried it on the GPU yet....and I will need to change the test to launch a kernel....but I think there should be support??
https://docs.nvidia.com/cuda/cuda-c-programming-guide/#assertion

artv3:
@rrsettgast , @davidbeckingsale , I tried out the branch and assert worked for me (on the GPU). One suggestion is to add the <cassert> header in ManagedArray.inl . I would add it myself but I don't have permissions to push onto this branch.

rrsettgast:
@artv3 do you mean including "assert.h" in ManagedArray.inl?

artv3:
right @rrsettgast,  ``assert.h`` or <cassert> would also work.

davidbeckingsale:
Are you sure `cassert` works?

artv3:
@davidbeckingsale , it did for me. I also used GNU-4.9 as my host compiler.

PR #35, opened 2018-06-08
Pick implementation in CHAI/Umpire

rfhaque:
Implementation of pick/set/incr/decr methods in CHAI

### DIFF
--- a/src/ArrayManager.inl
+++ b/src/ArrayManager.inl
@@ -52,6 +52,10 @@
 
 #include "umpire/ResourceManager.hpp"
 
+#if defined(CHAI_ENABLE_UM)
+#include <cuda_runtime_api.h>
+#endif
+
 namespace chai {
 
 template<typename T>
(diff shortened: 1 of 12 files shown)


src/ArrayManager.inl
```diff
@@ -119,6 +123,36 @@ void* ArrayManager::reallocate(void* pointer, size_t elems, PointerRecord* point
...
 
+#if defined(CHAI_ENABLE_PICK)
+template<typename T>
+CHAI_INLINE
+typename ArrayManager::T_non_const<T> ArrayManager::pick(T* src_ptr, size_t index)
+{
+  auto dst_alloc_strategy = m_resource_manager.getAllocator("HOST").getAllocationStrategy();
+  auto src_alloc_strategy = m_resource_manager.getAllocator(const_cast<T_non_const<T>*>(src_ptr)).getAllocationStrategy();
+  T_non_const<T> dst;
+  m_resource_manager.transfer(&dst /*Host*/,
```
davidbeckingsale: Why can't you use the `copy` method here?

rfhaque: One pointer in each case (`dst_ptr` in `pick` and `src_ptr` in `set`) is a non-umpire pointer. Hence the call to find the allocation record in the `copy` method would fail.

davidbeckingsale: Were you able to update this to remove the `transfer` requirement?

PR #39, opened 2018-06-18
Feature/CHAI-42-CHAI-Copyable

adayton1:
Added support for CHAI copyable. This ensures that inner data gets moved correctly in the situation where ManagedArrays are nested.

### DIFF
--- a/src/ManagedArray.hpp
+++ b/src/ManagedArray.hpp
@@ -57,6 +57,18 @@ namespace chai {
 struct InvalidConstCast;
 
 /*!
+ * \class CHAICopyable
+ *
+ * \brief Provides an interface for types that are copyable.
+ *
+ * If a class inherits from CHAICopyable, then a data transfer is triggered
+ * when the copy constructor is called.
+ */
+class CHAICopyable {
+   public:
+};
+
+/*!
  * \class ManagedArray
  *
  * \brief Provides an array-like class that automatically transfers data
@@ -73,7 +85,7 @@ struct InvalidConstCast;
  * \tparam T The type of elements stored in the ManagedArray.
  */
 template <typename T>
-class ManagedArray {
+class ManagedArray : public CHAICopyable {
   public:
 
   using T_non_const = typename std::remove_const<T>::type;
@@ -235,6 +247,12 @@ class ManagedArray {
   CHAI_HOST_DEVICE ManagedArray<T>& operator= (std::nullptr_t);
 
   CHAI_HOST_DEVICE bool operator== (ManagedArray<T>& rhs);
+
+  template<bool B = std::is_base_of<CHAICopyable, T>::value, typename std::enable_if<B, int>::type = 0>
+  CHAI_HOST_DEVICE void migrateInnerData();
+
+  template<bool B = std::is_base_of<CHAICopyable, T>::value, typename std::enable_if<!B, int>::type = 0>
+  CHAI_HOST_DEVICE void migrateInnerData();
 
   private:
 
--- a/src/ManagedArray.hpp
+++ b/src/ManagedArray.hpp
@@ -61,11 +61,10 @@ struct InvalidConstCast;
  *
  * \brief Provides an interface for types that are copyable.
  *
- * If a class inherits from CHAICopyable, then a data transfer is triggered
- * when the copy constructor is called.
+ * If a class inherits from CHAICopyable, then the stored items of type T
+ * are moved when the copy constructor is called.
  */
 class CHAICopyable {
-   public:
 };
 
 /*!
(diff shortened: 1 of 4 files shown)


src/ManagedArray.hpp
```diff
@@ -56,6 +56,18 @@ namespace chai {
...
+/*!
+ * \class CHAICopyable
+ *
+ * \brief Provides an interface for types that are copyable.
+ *
+ * If a class inherits from CHAICopyable, then a data transfer is triggered
+ * when the copy constructor is called.
+ */
+class CHAICopyable {
+   public:
```
davidbeckingsale: Is this line necessary?

src/ManagedArray.inl
```diff
@@ -110,6 +110,8 @@ CHAI_HOST_DEVICE ManagedArray<T>::ManagedArray(ManagedArray const& other):
 
   m_active_pointer = static_cast<T*>(m_resource_manager->move(const_cast<T_non_const*>(m_active_pointer), m_pointer_record));
 
+  migrateInnerData();
```
davidbeckingsale: Can we rename this, something like 'moveInnerData` or `copyInnerData`. "MIgrate" means we are mixing verbs.

src/ManagedArray.hpp
```diff
@@ -236,6 +248,12 @@ class ManagedArray {
 
   CHAI_HOST_DEVICE bool operator== (ManagedArray<T>& rhs);
 
+  template<bool B = std::is_base_of<CHAICopyable, T>::value, typename std::enable_if<B, int>::type = 0>
```
davidbeckingsale: Can you add comments clarifying when each implementation well be used.  

(I understand, but it would be helpful for future reference)

src/ManagedArray.hpp
```diff
@@ -56,6 +56,18 @@ namespace chai {
 
 struct InvalidConstCast;
 
+/*!
+ * \class CHAICopyable
+ *
+ * \brief Provides an interface for types that are copyable.
+ *
+ * If a class inherits from CHAICopyable, then a data transfer is triggered
```
davidbeckingsale: Might want to clarify that it triggers a copy each of the stored items of type `T`.

src/ManagedArray.hpp
adayton1: Done.

src/ManagedArray.hpp
adayton1: Done.

src/ManagedArray.inl
adayton1: Changed to "move".

src/ManagedArray.hpp
adayton1: Nope.

davidbeckingsale:
Looks good.

adayton1:
A lot of the tests are failing, including this one. It looks like an Umpire Exception.

`[ERROR][/g/g17/dayton8/ale3d/chai/build/include/umpire/alloc/CudaMallocAllocator.hpp:42]: allocate cudaMalloc( bytes = 96 ) failed with error: unknown error
unknown file: Failure
C++ exception with description "! Umpire Exception [/g/g17/dayton8/ale3d/chai/build/include/umpire/alloc/CudaMallocAllocator.hpp:42]:  allocatecudaMalloc( bytes = 96 ) failed with error: unknown error" thrown in the test body.`

davidbeckingsale:
Where are you running this?

adayton1:
I'm running it on rzmanta.

davidbeckingsale:
Are you running the tests on a compute node?

adayton1:
That would be the smart thing to do...

adayton1:
Any ideas on a good test case? I'm a little stuck.

src/tests/managed_array_tests.cpp
```diff
@@ -409,4 +409,40 @@ CUDA_TEST(ManagedArray, Move)
...
+
+  forall(cuda(), 0, 3, [=] __device__ (int i) {
+      gpuArray[0][i] = i;
+      gpuArray[1][i] = i;
+      gpuArray[2][i] = i;
+  });
+
+  auto hostArray = chai::ManagedArray<chai::ManagedArray<int>>(gpuArray);
+
+  for (int i = 0; i < 3; ++i)
```
davidbeckingsale: Make this loop look like the one above, but use sequential instead of cuda. You can have multiple ASSERT statements inside the loop.

adayton1: OK, I think I have a good test case now. I did notice that only shallow copies are being done currently. Is that correct?

PR #40, opened 2018-06-18
ENH: Added a deepCopy free function.

corbett5:


src/ManagedArray.hpp
```diff
@@ -297,6 +297,32 @@ makeManagedArray(
...
+deepCopy(
+    ManagedArray<T> const& array)
+{
+  T* data_ptr = array.getActivePointer();
+  
+  ArrayManager* manager = ArrayManager::getInstance();
+  
+  PointerRecord const* record = manager->getPointerRecord(data_ptr);
+  
+  PointerRecord* copy_record = manager->copyRecord(record);
```
davidbeckingsale: It make more sense to call this `copyFromRecord` or `deepCopyRecord` - otherwise it sounds like it just copies the record and not the data.

src/tests/managed_array_tests.cpp
```diff
@@ -410,3 +410,54 @@ CUDA_TEST(ManagedArray, Move)
...
+
+  chai::ManagedArray<float> copy = chai::deepCopy(array);
+  ASSERT_EQ(copy.size(), 10u);
+
+  forall(cuda(), 0, 10, [=] __device__ (int i) {
+      array[i] = 3.14159 * i;
+  });
+
+  forall(sequential(), 0, 10, [=] (int i) {
+    ASSERT_EQ(copy[i] = i);
```
davidbeckingsale: I think this should be `ASSERT_EQ(copy[i], i)`

You could also check that the original array is != to the copy.

src/ManagedArray.hpp
corbett5: Done.

src/tests/managed_array_tests.cpp
corbett5: I added a check that original array contains the updated values.

src/tests/managed_array_tests.cpp
```diff
@@ -410,3 +410,54 @@ CUDA_TEST(ManagedArray, Move)
...
+CUDA_TEST(ManagedArray, DeviceDeepCopy)
+{
+  chai::ManagedArray<float> array(10, chai::GPU);
+  ASSERT_EQ(array.size(), 10u);
+
+  forall(cuda(), 0, 10, [=] __device__ (int i) {
+      array[i] = i;
+  });
+
+  chai::ManagedArray<float> copy = chai::deepCopy(array);
```
corbett5: @davidbeckingsale So my original intent with the deepCopy function was to only copy the data in the active region. Ie the `copy` array would only have a device allocation. But by nature of using the copy constructor here the data will get copied back to the host correct?

corbett5: Actually now that I look at it a little more the `operator=(ManagedArray&)` isn't provided so I'm guessing this is doing an automatic conversion to `T*` then using the `ManagedArray(T*)` constructor. What do you think the best way to go about this is?

corbett5: Nevermind, I thought providing the `operator=(std::nullptr_t)` would disable the compiler generated assignment operator, but I guess not.

davidbeckingsale: So this is working as you expect?

corbett5: Yup!

davidbeckingsale:
@adayton1 - can you check this works with `CHAICopyable`?

PR #41, opened 2018-06-25
CHAI-34-leak-checker

adayton1:
Added a function to return the map of pointers for the purposes of leak checking.

### DIFF
--- a/src/ArrayManager.hpp
+++ b/src/ArrayManager.hpp
@@ -218,6 +218,13 @@ class ArrayManager
    */
   PointerRecord* deepCopyRecord(PointerRecord const* record);
 
+  /*!
+   * \brief Create a copy of the pointers map.
+   *
+   * \return A copy of the pointers map.
+   */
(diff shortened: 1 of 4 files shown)


src/ArrayManager.hpp
```diff
@@ -218,6 +218,13 @@ class ArrayManager
    */
   PointerRecord* deepCopyRecord(PointerRecord const* record);
 
+  /*!
+   * \brief Create a copy of the pointers map.
```
davidbeckingsale: pointers -> pointer.

You might want to put a comment here about the expected use case, e.g. to check on memory usage, find leaks, etc.

davidbeckingsale:
This looks good.

One thing I think would be really cool would be to code up a small example showing how this could be used to find a leak.

src/ArrayManager.hpp
adayton1: Done.

PR #42, opened 2018-06-26
Fixed bug in copyable

adayton1:
Previously used incorrect version of copyable functionality. Put in the correct version.


src/chai/ManagedArray.inl
```diff
@@ -378,21 +377,21 @@ template<bool B, typename std::enable_if<B, int>::type>
...
 void
-ManagedArray<T>::moveInnerData()
+ManagedArray<T>::moveInnerImpl()
 {
-   for (int i = 0; i < size(); ++i)
-   {
-      // Copy constructor triggers data movement
-      T temp = T(m_active_pointer[i]);
-   }
+  // Use reverse order to optimize cnmem deallocation
```
davidbeckingsale: I don't think this comment should be here anymore 😄

adayton1: You don't think it needs to be in reverse order anymore?

adayton1: Fixed!

PR #43, opened 2018-06-26
Added some functions for additional leak checking

adayton1:
Added functions to get the number of arrays registered with the resource manager and the total amount of memory allocated.


davidbeckingsale:
Why have you added these functions to the ArrayManager?

adayton1:
For convenience. Would you prefer not to have them on the ArrayManager?

PR #46, opened 2018-07-12
change the way the const conversion operator is defined

rrsettgast:


davidbeckingsale:
Your fix looks good. I'm just running it through some tests locally.

PR #47, opened 2018-07-17
ENH: Added move constructor and move assignment operator.

corbett5:


davidbeckingsale:
@ajkunen @robinson96 do you have any thoughts on this? It doesn't seem to impact the correctness of the copy semantics.

corbett5:
I've managed to work around the absence of the move constructor and move assignment operator in Geos so while they would still be nice to have there's no need for them.

PR #49, opened 2018-07-24
Add support for shared BLT outside of the repository

liu15:
Add support for shared BLT directory outside of the repository.
If BLT_SOURCE_DIR is set, then that directory will be used for BLT, otherwise, the internal one will be used.


davidbeckingsale:
Nice, thanks!

PR #55, opened 2018-10-11
Let ManagedArrays use different umpire::Allocators

davidbeckingsale:


davidbeckingsale:
@ajkunen - this is almost working, I'm running into some strange issues with the callbacks that I haven't tracked down yet.

Note that the initializer-list constructors don't work with `{{ }}` syntax for nvcc, so need to use the `std::initializer_list<T>{}` form ☹️

davidbeckingsale:
@mcfadden8 - I tracked down the bug, this is ready for review.

src/chai/ManagedArray.hpp
```diff
@@ -91,10 +94,14 @@ class ManagedArray : public CHAICopyable
...
 
+  CHAI_HOST_DEVICE ManagedArray();
+
   /*!
    * \brief Default constructor creates a ManagedArray with no allocations.
    */
-  CHAI_HOST_DEVICE ManagedArray();
+  CHAI_HOST_DEVICE ManagedArray(
+      std::initializer_list<chai::ExecutionSpace> spaces,
+      std::initializer_list<umpire::Allocator> allocators);
```
davidbeckingsale: @trws - I couldn't make these constructors work using double brace enclosed initializer lists with nvcc. Should I be able to make it work?

trws: I'm honestly not sure, would have to dig into this further.

davidbeckingsale: No problems, thanks for looking.

src/chai/ManagedArray.inl
```diff
@@ -58,21 +58,71 @@ CHAI_HOST_DEVICE ManagedArray<T>::ManagedArray():
...
-  m_elems(elems),
-  m_pointer_record(nullptr)
+    std::initializer_list<chai::ExecutionSpace> spaces,
+    std::initializer_list<umpire::Allocator> allocators):
+  ManagedArray()
 {
 #if !defined(__CUDA_ARCH__)
-  m_resource_manager = ArrayManager::getInstance();
+  int i = 0;
+  for (auto& space : spaces) {
```
davidbeckingsale: @trws - looking at this method, do you think there is a better way to implement this? I could use arrays instead - do you think that would be a good choice?

Note that this method is instantiated by user code, so we should be able to infer the array size as a template parameter.

trws: I like taking arrays, or at least a constant-sized thing, if it only makes sense to take a known number of items.  If this is meant to have dynamic rank, then I'd leave it as is, otherwise I'd probably switch it to something that enforces the restriction at compile time.

davidbeckingsale:
It’s designed to take 1 to N items, where N is known at compile time. I’ll see if I can make an array version work.

src/chai/ManagedArray.hpp
```diff
@@ -107,7 +115,16 @@ class ManagedArray : public CHAICopyable
...
    */
-  CHAI_HOST_DEVICE ManagedArray(size_t elems, ExecutionSpace space = NONE);
+  CHAI_HOST_DEVICE ManagedArray(
+      size_t elems,
+      ExecutionSpace space = NONE);
+
+  template<int N>
+  CHAI_HOST_DEVICE ManagedArray(
+      size_t elems,
+      std::array<chai::ExecutionSpace, N> spaces,
```
davidbeckingsale: @trws - this is failing to deduce `N` when I try and construct an array with a double brace initializer list. Any ideas?

trws:
That seems odd, I do not know.

On 19 Oct 2018, at 11:58, David Beckingsale wrote:

@davidbeckingsale commented on this pull request.

________________________________

In src/chai/ManagedArray.hpp<https://github.com/LLNL/CHAI/pull/55#discussion_r226749948>:

> @@ -107,7 +115,16 @@ class ManagedArray : public CHAICopyable
    * \param elems Number of elements in the array.
    * \param space Execution space in which to allocate the array.
    */
-  CHAI_HOST_DEVICE ManagedArray(size_t elems, ExecutionSpace space = NONE);
+  CHAI_HOST_DEVICE ManagedArray(
+      size_t elems,
+      ExecutionSpace space = NONE);
+
+  template<int N>
+  CHAI_HOST_DEVICE ManagedArray(
+      size_t elems,
+      std::array<chai::ExecutionSpace, N> spaces,


@trws<https://github.com/trws> - this is failing to deduce N when I try and construct an array with a double brace initializer list. Any ideas?

davidbeckingsale:
@ajkunen - any comments on the interface?

PR #63, opened 2019-02-01
Made the const semantics work with nested arrays.

corbett5:


corbett5:
@davidbeckingsale I tested it on quartz with clang 6 and ray with clang-coral. Sorry about last time, I forgot to test without nvcc.

davidbeckingsale:
@rrsettgast - can we get another review...

PR #68, opened 2019-03-22
Bugfix/corbett/memory leak

corbett5:
Related to LLNL/Umpire#84. Only merge after that PR and after updating umpire with the subsequent merge commit.

### DIFF
--- a/docs/doxygen/CMakeLists.txt
+++ b/docs/doxygen/CMakeLists.txt
@@ -41,4 +41,4 @@
 # POSSIBILITY OF SUCH DAMAGE.
 #######################################################################
 
-add_doxygen_target(chai_doxgen)
+blt_add_doxygen_target(chai_doxgen)
--- a/src/chai/ArrayManager.cpp
+++ b/src/chai/ArrayManager.cpp
@@ -81,6 +81,20 @@ ArrayManager::ArrayManager()
 #endif
 }
 
+void ArrayManager::finalize()
+{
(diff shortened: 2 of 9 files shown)


src/chai/ArrayManager.cpp
```diff
@@ -81,6 +81,20 @@ ArrayManager::ArrayManager()
 #endif
 }
 
+void ArrayManager::finalize()
```
corbett5: @davidbeckingsale whatever we decide to do in Umpire I'll do here as well.
https://github.com/LLNL/Umpire/pull/84#discussion_r267993850

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "blt"]
 	path = blt
-	url = https://github.com/LLNL/blt.git
+	url = ../blt.git
```
davidbeckingsale: Nope.

docs/doxygen/CMakeLists.txt
```diff
@@ -41,4 +41,4 @@
 # POSSIBILITY OF SUCH DAMAGE.
 #######################################################################
 
-add_doxygen_target(chai_doxgen)
+blt_add_doxygen_target(chai_doxgen)
```
davidbeckingsale: Is this because of a BLT update?

corbett5: Correct. I updated blt to the same commit as Umpire.

PR #69, opened 2019-03-27
Bugfix/corbett/some leaks

corbett5:


corbett5:
@davidbeckingsale would you mind giving this a review?

src/chai/ManagedArray.hpp
```diff
@@ -134,6 +134,8 @@ class ManagedArray : public CHAICopyable
    */
   CHAI_HOST_DEVICE ManagedArray(ManagedArray const& other);
 
+  CHAI_HOST_DEVICE ManagedArray(ManagedArray && other);
```
davidbeckingsale: What's this for?

src/chai/ManagedArray.hpp
```diff
@@ -226,6 +228,10 @@ class ManagedArray : public CHAICopyable
                                 size_t m_elems,
                                 PointerRecord* pointer_record);
 
+  ManagedArray<T>& operator=(ManagedArray const & other) = default;
+
+  CHAI_HOST_DEVICE ManagedArray<T>& operator=(ManagedArray && other);
```
davidbeckingsale: And this one?

src/chai/ManagedArray.hpp
corbett5: This along with the move assignment operator perform the default operation, but now also do `other = nullptr`. The reason for clearing `other` is so that if a wrapper class were then to call `other.free()` it wouldn't delete the pointer record now "owned" by `*this`.

davidbeckingsale: We didn't have a move operator before though?

corbett5: True, I forgot that. Is there a reason to not have a move constructor?

davidbeckingsale: We discussed this before, and decided not to add one (https://github.com/LLNL/CHAI/pull/47) in case it impacts any application correctness I thought.

corbett5: Wow, I can't believe I forgot that. My bad. Is it ok to keep the move assignment operator though? That would be a lot less disruptive. If not it's fine, you can mimic the move semantics with the following
```c++
ManagedArray<T> toMove(10);
ManagedArray<T> x(nullptr);
x = toMove;
toMove = nullptr;
```

Keeping the move assignment operator would condense the last two, so it would be nice but certainly not necessary.

davidbeckingsale: I would be fine with move assigment. The rest of this looks good.

corbett5: Done.

corbett5:
@davidbeckingsale All the travis tests are failing but it seems to be failing on `make install` due to a permission issue so I'm guessing it's not related to this PR.

https://travis-ci.org/LLNL/CHAI/jobs/524226031#L2034

davidbeckingsale:
Yup, @corbett5 don't worry about it. The azure tests are passing. I think I will drop Travis soon.

corbett5:
@davidbeckingsale if you don't have any other requests could you merge this?

PR #74, opened 2019-05-22
Added __CUDACC__ guard to host-device definitions.

corbett5:


src/chai/ChaiMacros.hpp
```diff
@@ -45,7 +45,7 @@
 
 #include "chai/config.hpp"
 
-#if defined(CHAI_ENABLE_CUDA)
+#if defined(CHAI_ENABLE_CUDA) && defined(__CUDACC__)
```
davidbeckingsale: Nice. Have you run into this as a problem somewhere?

corbett5: Right now in geos. I tried to get around it by adding cuda as a dependency to everything but that led to other issues. We do something similar in axom:
https://github.com/LLNL/axom/blob/6d8f3cf2ff523678bf20a9df69c4dabc6b028c31/src/axom/core/Macros.hpp#L28

davidbeckingsale: Makes sense to me, thanks 👍

corbett5:
@davidbeckingsale since the Azure tests passed can you merge this?

PR #84, opened 2019-09-27
Modified to support higher level projects with add_subdirectory

ajkunen:
This allows a higher-level project to bring in CHAI and Umpire directly, instead of requiring CHAI to bring in Umpire.


ajkunen:
yay!  You guys are going to have to merge, I don't have permission on this project

PR #85, opened 2019-10-01
Adds a portable pattern for polymorphism

adayton1:
The new managed_ptr class provides a uniform interface for using polymorphism on the host and the device. As some classes may contain nested ManagedArrays or managed_ptrs, callbacks can be used to assist with data movement if necessary for the nested CHAI critters.


davidbeckingsale:
@adayton1 - this is really cool, thanks for such a neat feature PR.

Can you try and fixup the builds (I think CPU-only builds are failing), and then we can look at getting this merged!

adayton1:
I fixed the cpu build error in benchmarks. It should now work!

PR #90, opened 2019-10-30
using ManagedArray::reallocate() on a default-constructed ManagedArray doesn't actually allocate at all

wrtobin:


tests/unit/managed_array_unit_tests.cpp
```diff
@@ -35,6 +35,25 @@ TEST(ManagedArray, SpaceConstructorCPU)
   array.free();
 }
 
+TEST(ManagedArray, ReallocDefaultConstructor)
+{
+  chai::ManagedArray<float> array;
+  array.reallocate(1);
+  float val = array[0];
```
wrtobin: This causes a segfault in the current implementation, as the ManagedArray doesn't actually allocate anything during the reallocate call.

wrtobin:
Note I haven't checked for all tests passing on Lassen yet -- since the machine is still down at the moment.

tests/unit/managed_array_unit_tests.cpp
robinson96: changes look reasonable to me.

src/chai/ArrayManager.inl
```diff
@@ -47,20 +47,26 @@ void* ArrayManager::reallocate(void* pointer, size_t elems, PointerRecord* point
   for (int space = CPU; space < NUM_EXECUTION_SPACES; ++space) {
     void* old_ptr = pointer_record->m_pointers[space];
 
-    if (old_ptr) {
-      pointer_record->m_user_callback(ACTION_ALLOC, ExecutionSpace(space), sizeof(T) * elems);
-      void* new_ptr = m_allocators[space]->allocate(sizeof(T)*elems);
+    pointer_record->m_user_callback(ACTION_ALLOC, ExecutionSpace(space), sizeof(T) * elems);
+    void* new_ptr = m_allocators[space]->allocate(sizeof(T)*elems);
```
robinson96: looks good

davidbeckingsale: This will cause reallocate to allocate a pointer in every space - I'm not sure that's what we want.

davidbeckingsale: @wrtobin  @robinson96  - I think this PR introduces an unexpected change in behavior, can you take a look?

robinson96: ooh - you're right David. 
Bill, the model should  preserve lazy allocation of device memory so that reallocations don't inadvertently initialize device memory if they haven't used gpu memory before.  At one point, we had an implementation of reallocate that also strode to only do a deep copy of the data on any memory space where the data was active.

PR #91, opened 2019-11-12
Changes needed for GEOSX.

corbett5:
Changes I made to chai in-order to use it in GEOSX. Fixes some trivial leaks and adds a move constructor to the `ManagedArray`.

Doesn't need to go in but thought I'd put these up here.

### DIFF
--- a/src/chai/ManagedArray.hpp
+++ b/src/chai/ManagedArray.hpp
@@ -181,6 +181,12 @@ public:
   T* getActiveBasePointer() const;
 
   /*!
+   * \brief get access to m_pointer_record
+   * @return a copy of m_pointer_record
+   */
+  PointerRecord const * getPointerRecord() const;
+
+  /*!
    * \brief
    *
    */
--- a/src/chai/ManagedArray.hpp
+++ b/src/chai/ManagedArray.hpp
@@ -98,6 +98,13 @@ public:
   CHAI_HOST_DEVICE ManagedArray(ManagedArray const& other);
 
   /*!
+   * \brief Move constructor, does not trigger data movement.
+   *
+   * \param other ManagedArray being moved from.
+   */
+  CHAI_HOST_DEVICE ManagedArray(ManagedArray && other);
+
+  /*!
    * \brief Construct a ManagedArray from a nullptr.
    */
   CHAI_HOST_DEVICE ManagedArray(std::nullptr_t other);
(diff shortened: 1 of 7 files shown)


src/chai/ManagedArray.inl
```diff
@@ -408,6 +422,13 @@ ManagedArray<T>::getActiveBasePointer() const
   return m_active_base_pointer;
 }
 
+template<typename T>
+PointerRecord const * 
+ManagedArray<T>::getPointerRecord() const
```
davidbeckingsale: What's the use case for this?

src/chai/ManagedArray.hpp
```diff
@@ -97,6 +97,13 @@ class ManagedArray : public CHAICopyable
    */
   CHAI_HOST_DEVICE ManagedArray(ManagedArray const& other);
 
+  /*!
+   * \brief Move constructor, does not trigger data movement.
+   *
+   * \param other ManagedArray being moved from.
+   */
+  CHAI_HOST_DEVICE ManagedArray(ManagedArray && other);
```
davidbeckingsale: @ajkunen @robinson96 - are you opposed to a move constructor in ManagedArray?

src/chai/ManagedArray.inl
corbett5: In our wrapper class when we reallocate we create a new `ManagedArray` and move the data into it manually because our wrapper class can hold non POD data. We'd also like to transfer the user-callback to the new class. Getting the pointer record from the `ArrayManager` won't work if the `ManagedArray` hasn't been allocated yet, and in our use cases we set the call back before doing allocation.

corbett5: https://github.com/GEOSX/cxx-utilities/blob/fb749fdb6acff24edff49ac1b79d5c6c13a77d68/src/src/ChaiBuffer.hpp#L111

src/chai/ManagedArray.inl
```diff
@@ -324,6 +338,11 @@ CHAI_INLINE
 CHAI_HOST
 void ManagedArray<T>::move(ExecutionSpace space)
 {
+  /* If this ManagedArray isn't associated with an allocation return immediately. */
+  if (m_pointer_record == nullptr) {
```
corbett5: We have nested `ManagedArrays` where some inner arrays are empty and others are assigned to using the shallow copy `operator=`. We initialize these inner arrays using the `ManagedArray(nullptr_t)` constructor since the default constructor allocates a `PointerRecord` that needs to be free'd. When you copy-construct or move the outer array it moves the inner arrays, some of which have not been assigned to (`m_pointer_record == nullptr`).

PR #96, opened 2020-03-06
Add pinned memory support to CHAI

davidbeckingsale:


src/chai/ArrayManager.cpp
```diff
@@ -137,16 +145,24 @@ void ArrayManager::move(PointerRecord* record, ExecutionSpace space)
...
   if (record->m_last_space == UM) {
     return;
   }
 #endif
 
-  if (space == record->m_last_space) {
+#if defined(CHAI_ENABLE_PINNED)
+  if (record->m_last_space == PINNED) {
+    if (space == CPU) {
+      cudaDeviceSynchronize();
```
robinson96: As written, I believe this will cause a synchronize per array. Can we add some state to the resource manager that remembers it's already done a sync that gets reset when there is any move to a GPU space?

src/chai/ArrayManager.cpp
```diff
@@ -137,16 +149,25 @@ void ArrayManager::move(PointerRecord* record, ExecutionSpace space)
...
 #endif
 
-  if (space == record->m_last_space) {
+#if defined(CHAI_ENABLE_PINNED)
+  if (record->m_last_space == PINNED) {
+    if (space == CPU && m_need_sync_for_pinned) {
+      m_need_sync_for_pinned = false;
+      cudaDeviceSynchronize();
+    }
     return;
```
robinson96: I'd add some documentation to the early returns in this explaining why no move is required  (I had to look at the full file context to figure out this was where we were circumventing the copy).

davidbeckingsale: Will do, thanks for the review.

src/chai/ArrayManager.cpp
```diff
@@ -137,16 +149,25 @@ void ArrayManager::move(PointerRecord* record, ExecutionSpace space)
...
   }
 
+  if (space == record->m_last_space) {
+    return;
+  }
+
 #if defined(CHAI_ENABLE_UM)
   if (record->m_last_space == UM) {
```
robinson96: doesn't UM have the same correctness requirements as PINNED from a synchronization perspective? Should we use the m_need_sync_for_pinned (perhaps renamed?) for UM as well?

examples/pinned.cpp
```diff
@@ -0,0 +1,76 @@
...
+// Copyright (c) 2016-19, Lawrence Livermore National Security, LLC and CHAI
+// project contributors. See the COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+//////////////////////////////////////////////////////////////////////////////
+#include "../src/util/forall.hpp"
+#include "chai/ManagedArray.hpp"
+
+#include <iostream>
+
```
robinson96: can we add an example / test that uses slices with pinned buffers, just to ensure that all works as expected?

PR #97, opened 2020-04-01
BLT configuration fixes.

robinson96:
Use proper library names, track Umpire's possible dependency on mpi with ENABLE_MPI.


cmake/thirdparty/SetupChaiThirdparty.cmake
```diff
@@ -7,11 +7,16 @@
...
-
+    if (ENABLE_MPI)
+      set(UMPIRE_DEPENDS mpi)
+    else()
+      set(UMPIRE_DEPENDS)
+    endif()
     blt_register_library(
       NAME umpire
       INCLUDES ${UMPIRE_INCLUDE_DIRS}
-      LIBRARIES umpire)
```
davidbeckingsale, line 14: Using the named CMake targets should work. What was the problem you were seeing?

robinson96: The issue here is possible inconsistencies when integrating with other packages that depend on umpire. For most other BLT packages, we've seen / adopted the convention that the FindUmpire.cmak should define what it's libraries are. If one package defines umpires libraries to be 'umpire', and then exports that dependency, and then another defines umpire's libraries to be whatever it has in FindUmpire.cmake, then this can lead to multiple entries on a link line (e.g. full/path/to/umpire.a and simply 'umpire', or I've also seen duplicate entries for a full path - which nvcc barfs on). By deferring to the FindUmpire.cmake definition of its libraries, an application can point cmake to it's own FindUmpire.cmake to guarantee consistency.

davidbeckingsale: Who's providing `FindUmpire.cmake` (because I know it isn't umpire). I would prefer this to use BLT to check whether a registered library named Umpire already exists. Would that work for you?

robinson96: Anyone building has the option to provide a FindUmpire.cmake (module mode of find_package). If none is available, it uses the config mode. It's possible my issues were merely due to inconsistencies. I'll try to adopt config mode throughout our system to see if that gets me to a working state. Stay tuned.

PR #98, opened 2020-04-02
adds ability to simulate GPU memory motion on a host-only build.

robinson96:


src/chai/ArrayManager.hpp
```diff
@@ -279,6 +279,14 @@ class ArrayManager
    */
   void deregisterPointer(PointerRecord* record);
 
+  /*!
+   * \brief set the allocator for an execution space.
+   *
+   * \param space Execution space to set the default allocator for.
+   * \param allocator The allocator to use for this space. Will be copied into chai.
+   */
+  void setAllocator(ExecutionSpace space, umpire::Allocator &allocator);
```
davidbeckingsale, line 288: Is there a reason this is added in this PR? It is unused.

robinson96: it gets used in how ALE3D sets up default allocators for spaces. Agree that it's tangentially related. I'm grouping a bit to make avoid conflicts down the road.

PR #99, opened 2020-04-02
Feature/dayton8/eviction

robinson96:
Adds an eviction capability.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -23,10 +23,6 @@ blt_add_executable(
   SOURCES chai-umpire-allocators.cpp
   DEPENDS_ON ${chai_umpire_example_depends})
 
-blt_add_test(
-  NAME managed_array_test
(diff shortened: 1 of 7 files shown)


examples/CMakeLists.txt
```diff
@@ -23,17 +23,10 @@ blt_add_executable(
   SOURCES chai-umpire-allocators.cpp
   DEPENDS_ON ${chai_umpire_example_depends})
 
-blt_add_test(
```
davidbeckingsale, line 26: Hmm, deleting tests?

robinson96: this target didn't exist when I tried to rebuild - maybe my workspace was hosed? Also it looks like this test exists in the integration tests, not here in examples...

robinson96: Well, fixed my build issue. But this is still a duplicate entry - the target is also is added in tests/integration and in fact it is being tested twice on develop.

PR #105, opened 2020-04-08
Give more info to callbacks

adayton1:
This PR updates the callback signature to take the (const) PointerRecord instead of the size. From the pointer record, the size can easily be regained. To that effect, I have moved calls to the callback so the pointer record always has the correct size and data (ALLOC after the allocation has happened, MOVE after the move has happened, and FREE before the deallocation has happened).

The more important effect of this is to allow functors and data encapsulation to be used more easily. See the example below. That functor encapsulates the data I need cleanly and allows me to easily update that data (in a lambda, I would need to have pointers to static global data).

```
class MemoryMotionLogger {
   public:
      void operator(const PointerRecord* record, Action action, ExecutionSpace space) {
         if (action == ACTION_MOVE) {
            std::string name = m_records_to_names[record];
            printf("%s: At location %s moved %lu bytes to space %i\n", name.c_str(), m_location.c_str(), record->m_size, space);
         }

      static void setLocation(std::string location) { m_location = location; }

      static void setName(const PointerRecord* record, std::string name) {
         m_records_to_names[record] = name;
      }

   private:
      static std::string location;
      static std::unordered_map<const PointerRecord*, std::string> m_records_to_names;
};
```

### DIFF
--- a/src/chai/ArrayManager.cpp
+++ b/src/chai/ArrayManager.cpp
@@ -71,8 +71,7 @@ void ArrayManager::registerPointer(
                    pointer << " already there.  Deleting abandoned pointer record.");
 
         PointerRecord *foundRecord = *(found_pointer_record_pair->second);
-
-        callback(foundRecord, ACTION_FOUND_ABANDONED, space, foundRecord->m_size);
+        callback(foundRecord, ACTION_FOUND_ABANDONED, space);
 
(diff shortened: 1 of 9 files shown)


corbett5:
I like it!

davidbeckingsale:
This looks good to me, I would like to get feedback from @ajkunen before merging.

PR #106, opened 2020-04-09
Add a global callback

adayton1:
This pull request adds a global callback that applies to all ManagedArrays. This can provide a greatly simplified interface to work with in many cases.

### DIFF
--- a/src/chai/ArrayManager.cpp
+++ b/src/chai/ArrayManager.cpp
@@ -333,6 +333,11 @@ void ArrayManager::setUserCallback(void* pointer, UserCallback const& f)
   pointer_record->m_user_callback = f;
 }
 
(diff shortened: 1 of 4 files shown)


src/chai/ArrayManager.cpp
```diff
@@ -333,6 +333,11 @@ void ArrayManager::setUserCallback(void* pointer, UserCallback const& f)
   pointer_record->m_user_callback = f;
 }
 
+void ArrayManager::setUserCallback(UserCallback const& f)
```
davidbeckingsale, line 336: Everything looks good, my only suggestion would be to consider renaming this to `setGlobalUserCallback` to make it obvious what it's doing.

adayton1: Done.

PR #109, opened 2020-04-18
Add method to get allocator

adayton1:


src/chai/ArrayManager.hpp
```diff
@@ -311,6 +311,13 @@ class ArrayManager
    */
   CHAISHAREDDLL_API void * frontOfAllocation(void * pointer);
 
+  /*!
+   * \brief Get the allocator for an execution space.
+   *
+   * \param space Execution space of the allocator to get.
+   * \param allocator The allocator for this space.
+   */
+  void getAllocator(ExecutionSpace space, umpire::Allocator &allocator);
```
davidbeckingsale, line 320: Is there a particular reason you take an Allocator reference instead of returning an Allocator?

adayton1: I was actually going to ask which you'd prefer. It would be more convenient to return the allocator so I'm fine with doing that. Are you OK with me returning the reference or would you like me to return by value?

davidbeckingsale: Return by value would be great. The `Allocator` object is very lightweight, since all it's really doing is holding a pointer to the underlying `AllocationStrategy`.

adayton1: OK, fixed.

PR #110, opened 2020-04-28
adds support for a PINNED Execution space.

robinson96:


src/chai/ArrayManager.cpp
```diff
@@ -209,17 +223,27 @@ void ArrayManager::move(PointerRecord* record, ExecutionSpace space)
...
 #endif
 
-  callback(record, ACTION_CAPTURED, space);
-
-  if (space == record->m_last_space) {
+#if defined(CHAI_ENABLE_PINNED)
+  if (record->m_last_space == PINNED) {
+    if (space == CPU && m_need_sync_for_pinned) {
+      m_need_sync_for_pinned = false;
+      cudaDeviceSynchronize();
```
davidbeckingsale, line 242: This should be if-defd for HIP/CUDA, maybe hide it in a `chai::synchronize()` call or something.

src/chai/ArrayManager.cpp
```diff
@@ -147,6 +155,12 @@ void ArrayManager::setExecutionSpace(ExecutionSpace space)
   CHAI_LOG(Debug, "Setting execution space to " << space);
   std::lock_guard<std::mutex> lock(m_mutex);
 
+#if defined(CHAI_ENABLE_PINNED)
+  if (chai::GPU == space) {
+    m_need_sync_for_pinned = true;
```
adayton1, line 160: How do you handle if the array is const on the device?

src/chai/ArrayManager.hpp
```diff
@@ -422,6 +422,10 @@ class ArrayManager
    * Whether or not to synchronize on device after every CHAI kernel.
    */
   bool m_device_synchronize = false;
+
+#if defined(CHAI_ENABLE_PINNED)
+  bool m_need_sync_for_pinned = false;
```
adayton1, line 427: Document variable.

src/chai/ManagedArray.hpp
```diff
@@ -367,6 +367,8 @@ class ManagedArray : public CHAICopyable
 #endif
   }
 
+  CHAI_HOST_DEVICE bool isSlice() { return m_is_slice;}
```
adayton1, line 370: Document method.

src/chai/ManagedArray.inl
```diff
@@ -393,16 +395,24 @@ void ManagedArray<T>::move(ExecutionSpace space) const
...
+    if (m_pointer_record->m_last_space == UM) {
+    } else
+#endif
+#if defined(CHAI_ENABLE_PINNED)
+    if (m_pointer_record->m_last_space == PINNED) {
+    } else 
+#endif
      if (!std::is_const<T>::value) {
        CHAI_LOG(Debug, "T is non-const, registering touch of pointer" << m_active_pointer);
        m_resource_manager->registerTouch(m_pointer_record);
```
adayton1, line 408: Doesn't pinned sort of require a touch? If we write on the device, don't we want to synchronize before it is used on the host again?

src/chai/ArrayManager.cpp
robinson96: I've added a chai::synchronize that should handle hip and cuda.

src/chai/ArrayManager.cpp
robinson96: The synchronize is only called when a move would normally occur for a non pinned MA, so all the same const semantics are preserved.

src/chai/ManagedArray.inl
robinson96: for pinned memory, we only need to track whether globally we need to synchronize. This branch takes the approach of the ArrayManager marking pinned memory as potentially needing a sync whenever a GPU space is being used, and the sync will only occur if a non-const pinned pointer is needed on the host. It's not quite an optimal case, e.g. - non -const pin memory used on GPU, host context, gpu context without use of that pointer, another host context with that pointer would then trigger another synchronize. That said, it's pretty close to optimal - it does not trigger a synchronize for every pointer, for instance. In practice this is working very well in our application.

src/chai/ArrayManager.hpp
robinson96: done.

src/chai/managed_ptr.hpp
```diff
@@ -153,7 +133,7 @@ namespace chai {
    ///       only be valid in the correct context. Take care when passing raw pointers
    ///       as arguments to member functions.
    ///    Be aware that CHAI checks every CUDA API call for GPU errors by default. To
-   ///       turn off GPU error checking, pass -DCHAI_ENABLE_GPU_ERROR_CHECKING=OFF as
+   ///       turn off GPU error checking, pass -DCHAI_ENABLE_CHAI_GPU_ERROR_CHECKING=OFF as
```
adayton1, line 136: I think this shouldn't have been changed.

robinson96: this is now fixed

adayton1:
That GPU simulation mode fix was from Ben Liu.

PR #118, opened 2020-05-15
Removing HEADERS_OUTPUT_SUBDIR, which won't be recognized by the new …

ptsuji:
Removing HEADERS_OUTPUT_SUBDIR before the update to the latest BLT.


davidbeckingsale:
Thanks @ptsuji 👍

PR #121, opened 2020-06-08
Fixes for chai::makeManagedArray

kanye-quest:
- Adds handling for calls to chai::makeManagedArray with `owned = false` and a pointer from an existing ManagedArray.
- Adds support for using makeManagedArray when DISABLE_RM is set
- Adds move assignment operator for thin ManagedArray implementation; resolves link error when assigning from a temporary ManagedArray (`array = ManagedArray<T>(...);`)


src/chai/ManagedArray.hpp
```diff
@@ -454,21 +454,36 @@ ManagedArray<T> makeManagedArray(T* data,
...
   if (!std::is_const<T>::value) {
     array.registerTouch(space);
   }
+#else
+  PointerRecord recordTmp;
+  recordTmp.m_pointers[space] = data;
+  recordTmp.m_size = sizeof(T) * elems;
+  recordTmp.m_owned[space] = owned;
+
+  ManagedArray<T> array(&recordTmp, space);
```
davidbeckingsale, line 485: You're passing a stack variable here as a pointer, I don't think this will work.

kanye-quest: This code creates the ManagedArray for the case where DISABLE_RM is enabled; the ManagedArray constructor in that case only copies over the values from the PointerRecord pointer to initialize m_active_pointer, m_elems, and m_is_slice (set to m_owned[space]), while m_pointer_record is set to null as in the other thin ManagedArray constructors.

src/chai/ManagedArray.hpp
```diff
@@ -454,21 +454,36 @@ ManagedArray<T> makeManagedArray(T* data,
...
+    // pointer has an owning PointerRecord, create a non-owning ManagedArray
+    // slice
+    array = array.slice(0, elems);
+  }
 
   if (!std::is_const<T>::value) {
     array.registerTouch(space);
   }
+#else
+  PointerRecord recordTmp;
```
adayton1, line 480: I just realized, I don't think we use pointer records when CHAI_DISABLE_RM is defined. They are usually set to nullptr.

kanye-quest: The PointerRecord gets used in the CHAI_DISABLE_RM codepath just to pass the pointers and slice/ownership status to the ManagedArray constructor; the underlying m_pointer_record field in the ManagedArray is still set to nullptr.

PR #125, opened 2020-06-30
fixes for ENABLE_IMPLICIT_CONVERSIONS=Off

dtaller:
In the future, ale3d plans to use chai with ENABLE_IMPLICIT_CONVERSIONS off. Having this option on (as is the case by default) has been the source of a bunch of gpu-related bugs in ale3d. When an int* is converted to a managed array implicitly without passing an array length, errors occur as data is copied to/from the gpu.

### DIFF
--- a/src/chai/managed_ptr.hpp
+++ b/src/chai/managed_ptr.hpp
@@ -664,7 +664,7 @@ namespace chai {
       ///
       template <typename T>
       CHAI_HOST_DEVICE T* getRawPointers(ManagedArray<T> arg) {
-         return (T*) arg;
+         return arg.getActivePointer();
       }
 
       ///
--- a/src/chai/managed_ptr.hpp
+++ b/src/chai/managed_ptr.hpp
@@ -664,7 +664,7 @@ namespace chai {
       ///
       template <typename T>
       CHAI_HOST_DEVICE T* getRawPointers(ManagedArray<T> arg) {
-         return arg.getActivePointer();
+         return arg.data();
       }
 
       ///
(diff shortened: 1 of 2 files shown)


src/chai/managed_ptr.hpp
```diff
@@ -664,7 +664,7 @@ namespace chai {
       ///
       template <typename T>
       CHAI_HOST_DEVICE T* getRawPointers(ManagedArray<T> arg) {
-         return (T*) arg;
+         return arg.getActivePointer();
```
adayton1, line 667: This is incorrect. `getActivePointer()` does not move data to the correct space. I have a pull request in that adds the "data" method that does the same thing as the cast. Once that is in, you can safely use that here. `return arg.data();`

tests/unit/array_manager_unit_tests.cpp
```diff
@@ -18,6 +18,8 @@ TEST(ArrayManager, Constructor)
 
 #ifndef CHAI_DISABLE_RM
 
+#if defined(CHAI_ENABLE_IMPLICIT_CONVERSIONS)
```
adayton1, line 21: This test shouldn't be relying on implicit conversions. Once my other pull request goes in, use the data method.

src/chai/managed_ptr.hpp
dtaller: Hi @adayton1 . Should be fixed

tests/unit/array_manager_unit_tests.cpp
dtaller: Done, fixed

PR #126, opened 2020-07-01
Adds a data method to use instead of implicit casts

adayton1:


adayton1:
@corbett5 Does this work with the wrapper you have in GEOS?

adayton1:
It looks like GEOSX uses a wrapper instead of inheriting from chai::ManagedArray. I don't think these changes will have any impact, there.

corbett5:
@adayton1 Yeah we use the `chai::ArrayManager` not `chai::ManagedArray` so this won't effect us at all.

PR #128, opened 2020-07-15
Removeimplicitcast2: fixes for ENABLE_IMPLICIT_CONVERSIONS=Off

dtaller:
I had some issues with git on my last pull request, was not able to check things out, detached head state.

This branch also has the fixes for disabling implicit casts


davidbeckingsale:
This seems identical to #125 ?

dtaller:
> This seems identical to #125 ?

Yes, I'm sorry. I had some trouble with git after you had merged in develop. After you merged in develop, a new build failure showed up on the nvcc continuous integration build, and I was not able to get to the bottom of it. The failure is internal to one of the libraries that was updated, but I was not clear on how to fix it. I'll think about it

PR #134, opened 2020-07-24
fix issue with make_managed with implicit conversions off

dtaller:
Previously, make_managed would fail when implicit conversion were turned off. This occurred because calling getRawPointers with host_device_ptrs would call this version of getRawPointers:

CHAI_HOST_DEVICE T getRawPointers(T arg) {
         return arg;
      }

Instead of this one:

      template <typename T>
      CHAI_HOST_DEVICE T* getRawPointers(ManagedArray<T> arg) {
         return arg.data();
      }

which is what we need to return a double* or int*

I fixed the issue by adding a another getRawPointers for host_device_ptr to assure that the compiler does what I want

### DIFF
--- a/src/chai/managed_ptr.hpp
+++ b/src/chai/managed_ptr.hpp
@@ -60,6 +60,12 @@
 #include <functional>
 
 
+// forward declaration of care::host_device_ptr
+namespace care {
+    template <typename T>
+    class host_device_ptr;
+}//namespace care
+
 namespace chai {
    namespace detail {
 #ifdef __CUDACC__
@@ -666,6 +672,24 @@ namespace chai {
       CHAI_HOST_DEVICE T* getRawPointers(ManagedArray<T> arg) {
          return arg.data();
       }
+
+      ///
+      /// @author Danny Taller
+      ///
+      /// This implementation of getRawPointers handles the CARE host_device_ptr type.
+      /// Note that without this function, care host_device_ptrs will go to the The non-CHAI type
+      /// version of this function (NOT the ManagedArray version), so the make managed paradigm
+      /// will fail unless implicit conversions are allowed.
+      ///
+      /// @param[in] arg The host_device_ptr from which to extract a raw pointer
+      ///
+      /// @return arg cast to a raw pointer
+      ///
+      template <typename T>
+      CHAI_HOST_DEVICE T* getRawPointers(care::host_device_ptr<T> arg) {
+         return ((ManagedArray<T>)arg).data();
+      }
+ 
 
       ///
       /// @author Alan Dayton


src/chai/managed_ptr.hpp
```diff
@@ -667,6 +673,24 @@ namespace chai {
...
+      /// Note that without this function, care host_device_ptrs will go to the The non-CHAI type
+      /// version of this function (NOT the ManagedArray version), so the make managed paradigm
+      /// will fail unless implicit conversions are allowed.
+      ///
+      /// @param[in] arg The host_device_ptr from which to extract a raw pointer
+      ///
+      /// @return arg cast to a raw pointer
+      ///
+      template <typename T>
+      CHAI_HOST_DEVICE T* getRawPointers(care::host_device_ptr<T> arg) {
```
davidbeckingsale, line 689: Can this specialization be added to care instead? I would rather not have a bi-directional dependency between the two projects.

dtaller: Yes, I know that this is very ugly. I tried adding this function on the care side already, but got the following error:

In file included from /usr/workspace/wsrzc/taller1/ALE3D/CHAI_IMPLICIT_CONSTR/exports/care/src/care/PointerTypes.h(21),
                 from /usr/workspace/wsrzc/taller1/ALE3D/CHAI_IMPLICIT_CONSTR/exports/care/src/care/care.h(267),
                 from /usr/workspace/wsrzc/taller1/ALE3D/CHAI_IMPLICIT_CONSTR/exports/care/src/care/LoopFuser.cpp(16):
/usr/workspace/wsrzc/taller1/ALE3D/CHAI_IMPLICIT_CONSTR/exports/care/src/care/host_device_ptr.h(434): error: function template "chai::detail::getRawPointers" has already been defined
        CHAI_HOST_DEVICE T* getRawPointers(host_device_ptr<T>& arg) {

It caused a conflict.

This is clearly ugly, but I had trouble coming up with a better solution. If you would like, we could wait until Monday and then see what @robinson96 or @adayton1 think when they are back, maybe they have a better idea.

davidbeckingsale: @dtaller - did you make any progress on alternative solutions? We are planning a CHAI release soon and it would be good to get in all the features you need.

dtaller: > @dtaller - did you make any progress on alternative solutions? We are planning a CHAI release soon and it would be good to get in all the features you need.

Hi David. Yes, I managed to put this function within CARE instead. Once I got the namespaces correct it worked. I think previously I might have messed that up.

But yes, we can close this pull request, it has been resolved by adding this getRawPointers function within CARE but within the the chai details namespace

davidbeckingsale: Great!

PR #137, opened 2020-07-28
Bump RAJA to v0.12.0 and update plugin

MrBurmark:
Update the CHAI RAJA plugin to work correctly with the changed RAJA plugin interface.
PreCapture and PostCapture calls have been added and the CHAI RAJA plugin now uses these instead of PreLaunch and PostLaunch.
The plugin tests have also been cleaned up a bit.


davidbeckingsale:
Thanks for this @MrBurmark - I am going to wait until the RAJA release comes out so that we can point at a tagged version in the submodule.

MrBurmark:
The pluginLinker in RAJA and pluginLinker in CHAI now conflict, suggestions?

davidbeckingsale:
> The pluginLinker in RAJA and pluginLinker in CHAI now conflict, suggestions?

Namespacing? Although the pluginLinker is in an anonymous namespace so it shouldn't conflict should it?

MrBurmark:
I think the problem comes when both headers are included in the same file. I'll add namespacing inside the anonymous namespace.

davidbeckingsale:
LGTM

PR #139, opened 2020-08-04
Fix: libcpp and ci

adrienbernede:
A minor fix to remove libcpp mentions in CI and fix the capture of the spec string in the CI jobs.

The two are related, since fixing the spec string (to allow whitespaces) uncovered the fact that there is no libcpp variant in CHAI package.

Decision was that there is not need for libcpp variant in CHAI at the moment.

EDIT: Also add a shared variant to CHAI package.


adrienbernede:
LGTM

adrienbernede:
@davidbeckingsale I’m not allowed to merge this. Would you please?

PR #142, opened 2020-08-10
Allow makeManaged to use a nullptr (returning the null record) with u…

liu15:
…mpire 3 (which does not accept nullptr registration)


liu15:
Sorry, I had another change from my PR on our project.

PR #146, opened 2020-09-02
test and example fixes when implicit casts are disabled

dtaller:


dtaller:
> I think it would be better to have these changes ifdeffed.

No worries, done.

davidbeckingsale (approved):
Awesome, this looks good. I will get it rolled into the release branch.

PR #165, opened 2020-10-29
Add support for GPU simulation mode in the RAJA plugin

adayton1:
Since RAJA currently does not have any notion of a GPU simulation mode, the "GPU kernels" will always run on the host and that is the platform that will be passed to the RAJA plugin. This pull request provides the capability for telling CHAI when GPU simulation mode is on or off. When it is off and the CAMP platform is "host", the execution space is set to chai::CPU. However, when it is on and the CAMP platform is "host", the execution space is set to chai::GPU. This works because the chai::GPU space is actually backed with host memory when configured with -DENABLE_GPU_SIMULATION_MODE=ON.

### DIFF
--- a/src/chai/ArrayManager.hpp
+++ b/src/chai/ArrayManager.hpp
@@ -425,7 +425,7 @@ public:
   /*!
    * \brief Return true if GPU simulation mode is on, false otherwise.
    */
-  bool getGPUSimMode() { return m_gpu_sim_mode; }
+  bool isGPUSimMode() { return m_gpu_sim_mode; }
 #endif
 
   /*!
--- a/src/chai/ArrayManager.hpp
+++ b/src/chai/ArrayManager.hpp
@@ -416,6 +416,18 @@ public:
    */
   bool syncIfNeeded();
 
+#if defined(CHAI_ENABLE_GPU_SIMULATION_MODE)
+  /*!
+   * \brief Turn the GPU simulation mode on or off.
+   */
+  void setGPUSimMode(bool gpuSimMode) { m_gpu_sim_mode = gpuSimMode; }
+
+  /*!
+   * \brief Return true if GPU simulation mode is on, false otherwise.
+   */
+  bool getGPUSimMode() { return m_gpu_sim_mode; }
+#endif
+
   /*!
    * \brief Evicts the data in the given space.
    *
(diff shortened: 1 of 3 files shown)


src/chai/ArrayManager.hpp
```diff
@@ -416,6 +416,18 @@ class ArrayManager
    */
   bool syncIfNeeded();
 
+#if defined(CHAI_ENABLE_GPU_SIMULATION_MODE)
+  /*!
+   * \brief Turn the GPU simulation mode on or off.
+   */
+  void setGPUSimMode(bool gpuSimMode) { m_gpu_sim_mode = gpuSimMode; }
```
davidbeckingsale, line 423: What's this used for?

src/chai/ArrayManager.hpp
```diff
@@ -416,6 +416,18 @@ class ArrayManager
...
+
+  /*!
+   * \brief Return true if GPU simulation mode is on, false otherwise.
+   */
+  bool getGPUSimMode() { return m_gpu_sim_mode; }
```
davidbeckingsale, line 428:
```suggestion
  bool inGpuSimMode() { return m_gpu_sim_mode; }
```

I'd rather this be an "is" or "is" method, I think it's more obvious it's returning a boolean.

davidbeckingsale (changes requested):

src/chai/ArrayManager.hpp
adayton1: CARE has a layer on top of RAJA. Calling setGPUSimMode before and after the call to RAJA is the only way I could think of to use the CHAI plugin for RAJA and still provide support for the GPU simulation mode. It would happen before and after this call: https://github.com/LLNL/CARE/blob/develop/src/care/forall.h#L78. I'm hoping this is only a temporary solution as RAJA seemed open to the idea of having a GPU simulation policy (but I doubt it will happen for a while).

src/chai/RajaExecutionSpacePlugin.cpp
```diff
@@ -25,13 +25,25 @@ RajaExecutionSpacePlugin::preCapture(const RAJA::util::PluginContext& p)
 
   switch (p.platform) {
     case RAJA::Platform::host:
-      m_arraymanager->setExecutionSpace(chai::CPU); break;
+#if defined(CHAI_ENABLE_GPU_SIMULATION_MODE)
```
robinson96, line 28: would this be cleaner if the logic got pushed into setExecutionSpace?

davidbeckingsale: That's a good idea.

src/chai/ArrayManager.hpp
adayton1: Fixed.

src/chai/RajaExecutionSpacePlugin.cpp
adayton1: Done.

davidbeckingsale:
LGTM

davidbeckingsale (approved):

PR #167, opened 2020-11-02
CHAI_LOG only active with CHAI_DEBUG regardless of RM being enabled

robinson96:
Not sure if this was intentional or not, but CHAI_LOG has been incurring overhead for non debug builds. This has shown up in some of CARE's stress testing kernel launch overhead (can double the cost of a kernel capture and launch). This PR changes CHAI_LOG to only do something when CHAI_DEBUG is defined.

### DIFF
--- a/src/chai/ChaiMacros.hpp
+++ b/src/chai/ChaiMacros.hpp
@@ -70,23 +70,21 @@
 
 #define CHAI_UNUSED_ARG(X)
 
+#if defined(CHAI_DEBUG)
 #if !defined(CHAI_DISABLE_RM)
 
 #define CHAI_LOG(level, msg) \
   UMPIRE_LOG(level, msg);
-
 #else
-
-#if defined(CHAI_DEBUG)
 
 #define CHAI_LOG(level, msg) \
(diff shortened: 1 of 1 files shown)


davidbeckingsale:
LGTM

davidbeckingsale:
I would rather this be a runtime option, because otherwise people will compile with it turned off and it makes it difficult to help debug issues.

Can we maybe remove the messages you were seeing cause a lot of overhead?

davidbeckingsale:
@adayton1 I think your PR #195 is similar to this, and I prefer the fine-grained approach here.

src/chai/ChaiMacros.hpp
```diff
@@ -70,23 +70,33 @@
 
 #define CHAI_UNUSED_ARG(X)
 
+#if !defined(CHAI_DISABLE_ALL_LOGGING)
 #if !defined(CHAI_DISABLE_RM)
```
liu15, line 74: Should this still be disabled if the resource manager is disabled?  Umpire is required either way, right?

PR #171, opened 2020-12-03
make tests and benchmarks workable for both cuda and hip

dtaller:
Replaced cuda specific guards like __CUDACC__ with CHAI_GPUCC (which is __CUDACC__ or __HIPCC__).

I used wrappers like chai::gpuMalloc instead of cuda-specific functions like cudaMalloc so that the code would things would compile for both architectures. Note that this has the added benefit (or side affect?) of adding error checking for these gpu calls:

`// wrapper for hip/cuda malloc`
`CHAI_HOST inline void gpuMalloc(void** devPtr, size_t size) {`
`#if defined (CHAI_ENABLE_HIP)`
`   CHAI_GPU_ERROR_CHECK(hipMalloc(devPtr, size));`
`#elif defined (CHAI_ENABLE_CUDA)`
`   CHAI_GPU_ERROR_CHECK(cudaMalloc(devPtr, size));`
`#endif`
`}`

### DIFF
--- a/benchmarks/chai_managed_ptr_benchmarks.cpp
+++ b/benchmarks/chai_managed_ptr_benchmarks.cpp
@@ -9,6 +9,7 @@
 #include "benchmark/benchmark.h"
 
 #include "chai/config.hpp"
+#include "chai/ArrayManager.hpp"
 #include "chai/managed_ptr.hpp"
 
 #include "../src/util/forall.hpp"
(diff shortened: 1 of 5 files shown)


adayton1 (approved):
LGTM

davidbeckingsale:
LGTM

robinson96 (approved):
LGTM

davidbeckingsale (approved):

PR #172, opened 2020-12-04
minor hip fix

dtaller:
Minor fix in the raja execution space logic. I am not using/testing this now for my own work, but I want it to be safe in case it becomes relevant in the future.

Still need to build on both cuda/hip; don't review yet!!!


adayton1:
Superseded by #177

PR #173, opened 2020-12-15
Add the ability to enable or disable managed_ptr

adayton1:
* Adds a config option for controlling whether managed_ptr support is enabled/disabled.


robinson96 (commented):
Do we want to keep managed_ptr in the includes so applications do not need to protect their include of managed_ptr with #ifdefs? Maybe with a #warning/#error to let people know they haven't protected their #include statement and an explicit naming of the appropriate macro to use so the compiler warning / error is more friendly?

adayton1:
Good idea, Peter. I just added a `#error` when managed_ptr.hpp is included and it is not supported. Let me know what you think.

robinson96 (approved):
LGTM

davidbeckingsale (approved):
👍

PR #176, opened 2021-03-16
prevent 0 byte allocations in ManagedArray_thin.incl

robinson96:


robinson96:
This is consistent with the behavior of ManagedArray.inl

davidbeckingsale (approved):

PR #193, opened 2021-08-19
Add initial HIP CI builds

davidbeckingsale:


davidbeckingsale:
Added to Azure in #208

PR #197, opened 2021-09-14
Fix hip build with resource manager disabled

adayton1:
HIP requires that both function declarations and definitions must have the same `__host__` and/or `__device__` specifiers.


davidbeckingsale:
Thanks Alan, our CI broke for an unrelated reason - fixing in #198 and then will merge this.

adayton1:
I doubted that the CI would have been broken by this :)

davidbeckingsale (approved):

PR #202, opened 2021-11-12
Add CHAI_ prefix and cmake dependent options

kab163:
Updating the cmake_dependent_options so that those cmake flags that are unique to CHAI will have the CHAI_ prefix. I also updated the azure_pipelines file so that it works properly with other Docker updates


davidbeckingsale (approved):
This looks good to me, thanks @kab163 !

kab163:
@davidbeckingsale this should be merged! (I don't have permissions lol)

PR #209, opened 2022-03-25
Add CMake option for initializing all memory to 0

adayton1:
In some cases, it is desirable to have memory initialized to 0 (particularly for debugging). This option is controlled by passing CHAI_ENABLE_ZERO_INITIALIZED_MEMORY=ON to CMake. It is off by default to avoid imposing any overhead.


rrsettgast:
Would it be also useful to be able to initialize values to like the maximum of the range so you would trigger an overflow if not initialized?

adayton1:
Are you thinking something like std::numeric_limits<T>::max()? I like the idea, but it does get a little more complicated since the particular type is not available for the ArrayManager::allocate (it appears to be available for ArrayManager::reallocate, however). So either we would have to add the type for the ArrayManager::allocate call, or we would have to plumb this into the ManagedArray class itself. @davidbeckingsale, any ideas?

PR #218, opened 2022-11-14
Woptim/update deps

adrienbernede:
As a consequence of local package update in Umpire and RAJA dependencies, we also update CHAI local packages to be in sync.

Relates to:
https://github.com/LLNL/Umpire/pull/793.
https://github.com/LLNL/RAJA/pull/1338

### DIFF
--- a/.gitlab/custom-jobs-and-variables.yml
+++ b/.gitlab/custom-jobs-and-variables.yml
@@ -19,7 +19,7 @@ variables:
 # Arguments for job level allocation
   RUBY_BUILD_AND_TEST_JOB_ALLOC: "--overlap --time=20 --nodes=1"
 # Project specific variants for ruby
-  PROJECT_RUBY_VARIANTS: ""
+  PROJECT_RUBY_VARIANTS: "~shared +raja tests=basic "
 # Project specific deps for ruby
   PROJECT_RUBY_DEPS: ""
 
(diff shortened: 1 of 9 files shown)


adrienbernede:
@kab163 Some cmake_dependent_options (`CMakeList.txt`) are also defined in `cmake/SetupChaiOptions.cmake`.
Looking at Umpire, I was under the impression that this was not necessary (and maybe source of confusion).

Do you know more about it ?

adrienbernede:
@davidbeckingsale re:

> Added cuda specs with “recent” toolchains (based on RAJA CI). 
> -> Failures in CHAI with recent CUDA. 
>      https://lc.llnl.gov/gitlab/radiuss/CHAI/-/jobs/907108#L505
>       https://lc.llnl.gov/gitlab/radiuss/CHAI/-/jobs/907107#L554

I added those specs only because CHAI had no spec actually running/passing with cuda on lassen. But we may ignore those to get the PR merge.

The failures in the corona tests may be due to the "old" rocm version. Testing now.

adrienbernede:
re:
> The failures in the corona tests may be due to the "old" rocm version.

It turns out the build fails with newer rocm on corona, while it passes on tioga.
The compiler is the same (clang@13), but on corona I get:

```
lld: error: undefined symbol: __oclc_ABI_version
```

Googling it, I see people saying they use the rocm provided clang to work around this. Could the tioga clang be installed differently than corona clang ?

PR #223, opened 2023-04-10
MigrateInner on CHAI Copyable for PINNED / UM

robinson96:
Fixes an issue where ManagedArray<CHAICopyable<T>> would not work as expected if the outer array was allocated in the PINNED execution space and the inner CHAICopyable values were allocated in CPU / GPU Execution spaces.


adayton1 (approved):
LGTM

PR #228, opened 2023-07-11
Deprecate API for synchronization after device kernels

adayton1:
* Deprecates ArrayManager::enableDeviceSynchronization and ArrayManager::disableDeviceSynchronization
* Instead, use the environment variables for device synchronization after all kernels (e.g. CUDA_LAUNCH_BLOCKING or HIP_LAUNCH_BLOCKING)

### DIFF
--- a/src/chai/ArrayManager.hpp
+++ b/src/chai/ArrayManager.hpp
@@ -399,16 +399,19 @@ public:
   /*!
    * \brief Turn on device synchronization after every kernel.
    */
+  [[deprecated("Set the CUDA_LAUNCH_BLOCKING or HIP_LAUNCH_BLOCKING environment variable instead")]]
   void enableDeviceSynchronize() { m_device_synchronize = true; }
 
   /*!
    * \brief Turn off device synchronization after every kernel.
    */
+  [[deprecated("Set the CUDA_LAUNCH_BLOCKING or HIP_LAUNCH_BLOCKING environment variable instead")]]
   void disableDeviceSynchronize() { m_device_synchronize = false; }
 
   /*!
    * \brief Turn on device synchronization after every kernel.
    */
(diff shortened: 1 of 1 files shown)


src/chai/ArrayManager.hpp
```diff
@@ -399,11 +399,13 @@ class ArrayManager
   /*!
    * \brief Turn on device synchronization after every kernel.
    */
+  [[deprecated("Set the CUDA_LAUNCH_BLOCKING or HIP_LAUNCH_BLOCKING environment variable instead")]]
   void enableDeviceSynchronize() { m_device_synchronize = true; }
```
dtaller, line 403: Hi Alan. This might seem very basic, but I have a few questions
1) What are the advantages of using an environment variable instead of this local variable? Couldn't you run into issues unexpected behavior at runtime with two users running the same code but within a different environment?
2) In C (and C++), there are functions that let you set environment variables. For example, you could do something like setenv("CUDA_LAUNCH_BLOCKING", "1", 1) ; . Would that make sense in this context? Make this function set the environment variable so things are consistent regardless?

adayton1: To give some broader context, CUDA and Hip kernels are launched asynchronously, so if there are errors, they may not be reported exactly when they happen. The standard way of debugging this kind of issue is to use the environment variables provided by CUDA and Hip to make kernels execute when they are called and make the CPU wait for them to finish before continuing. I was unaware of these environment variables when I added code in CHAI to explicitly synchronize after kernels in managed_ptr.hpp. That code is invasive and there's no real benefit to it, so I'd like to just remove it. But I figured it made more sense to deprecate it first and then after the next release remove it all together.

Regarding setting the environment variable in C and C++, I don't think it's worth it. Disabling it in particular could be really confusing to people who are using the environment variable since it would turn off synchronizes everywhere, not just in CHAI.

robinson96 (approved):
LGTM

dtaller (approved):

PR #229, opened 2023-07-13
setExecutionSpace thread safety

neelakausik:
Made m_current_execution_space and m_synced_since_last_kernel thread_local to address thread safety concerns.

### DIFF
--- a/src/chai/ArrayManager.cpp
+++ b/src/chai/ArrayManager.cpp
@@ -17,7 +17,7 @@
 namespace chai
 {
 thread_local ExecutionSpace ArrayManager::m_current_execution_space;
-thread_local bool ArrayManager::m_synced_since_last_kernel;
+thread_local bool ArrayManager::m_synced_since_last_kernel = false;
 
 PointerRecord ArrayManager::s_null_record = PointerRecord();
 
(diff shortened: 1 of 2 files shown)


src/chai/ArrayManager.cpp
```diff
@@ -32,7 +34,7 @@ ArrayManager::ArrayManager() :
   m_callbacks_active{true}
 {
   m_pointer_map.clear();
-  m_current_execution_space = NONE;
+  ArrayManager::m_current_execution_space = NONE;
```
adayton1, line 37: Adding ArrayManager:: here shouldn't be necessary.

adayton1 (dismissed):
LGTM

src/chai/ArrayManager.cpp
adayton1: Or inside any ArrayManager methods.

adayton1 (approved):
LGTM

dtaller:
Looks good to me.

Was there a failing test or a something like that that prompted this change, or were you just inspecting the code for possible issues?

dtaller (approved):

adayton1:
> Looks good to me.
> 
> Was there a failing test or a something like that that prompted this change, or were you just inspecting the code for possible issues?

@dtaller, we've been looking at launching kernels in separate GPU streams from separate CPU threads, which has led us to look into thread safety in CHAI and RAJA. This was one of the main issues that was uncovered.

davidbeckingsale:
Merged in #230

PR #235, opened 2023-08-21
Remove optional explicit synchronizes

adayton1:
* Removes ArrayManager::enableDeviceSynchronization and ArrayManager::disableDeviceSynchronization
* Instead, use the environment variables for device synchronization after all kernels (e.g. CUDA_LAUNCH_BLOCKING or HIP_LAUNCH_BLOCKING)

Follow up to #228


dtaller (approved):
Looks good to me.

Just check (email code leads) if there are any client codes that still use the depreciated functions, and let them know about this if needed.

adayton1:
> Looks good to me.
> 
> Just check (email code leads) if there are any client codes that still use the depreciated functions, and let them know about this if needed.

I got the deprecated message in before the release and waited until after the release to push this change in. That way codes will have a whole release cycle before being impacted by this. That being said, this was only for debugging and I don't believe anyone besides us is using this.

dtaller:
> > Looks good to me.
> > Just check (email code leads) if there are any client codes that still use the depreciated functions, and let them know about this if needed.
> 
> I got the deprecated message in before the release and waited until after the release to push this change in. That way codes will have a whole release cycle before being impacted by this. That being said, this was only for debugging and I don't believe anyone besides us is using this.

Good to hear, sounds good

robinson96 (approved):
LGTM

adayton1:
@davidbeckingsale, can you take a review pass and merge if you think this is ready?

davidbeckingsale (approved):

PR #242, opened 2024-01-29
Support for a single memory space writeable by both GPU and CPU

robinson96:
SUMMARY

Provides appropriate synchronizations when needed instead of performing any memory copies.

DETAILS

The configuration variable CHAI_THIN_GPU_ALLOCATE, when enabled in combination with CHAI_DISABLE_RM set to True, will configure CHAI to only perform necessary synchronization in lieu of necessary memory motion. 

Changes ManagedArray::m_size to ManagedArray::m_elems - now tracks #of bytes instead of number of entries in the array. This makes the ManagedArray more robust to type conversions.


davidbeckingsale:
This looks fine, but can you remove the tpl export change? We are going to bump to latest BLT and use the new TPL macros.

adayton1:
> This looks fine, but can you remove the tpl export change? We are going to bump to latest BLT and use the new TPL macros.

Fixed.

adayton1:
It sounds like the CZ is back up now, so is there an easy way to retrigger the pipelines other than adding a commit?

liu15 (approved):

PR #276, opened 2024-09-09
Fix syntax error in raja-chai integration test

liu15:
* Fix syntax error in tests when implicit conversions are disabled
* Avoid building non-essential stuff in umpire and raja (to speed up CI and also avoid running out of disk space in the CI)


adayton1 (approved):
LGTM

dtaller (approved):

PR #277, opened 2024-09-10
Add synchronizes before/after managed_ptr kernels

adayton1:
managed_ptr kernels currently run in stream 0, which is different than the default stream used by RAJA. To prevent any cross stream race conditions, we synchronize before and after those kernels. In the future, we will likely use the same default stream used by RAJA and/or modify the API such that there is an overload that takes a camp resource or stream as an argument.


liu15 (approved):

robinson96 (approved):
LGTM

MrBurmark:
I thought the streams created by camp should be synchronous with stream 0.

adayton1:
> I thought the streams created by camp should be synchronous with stream 0.

Is that true for both cuda and hip?

MrBurmark:
I know its true for cuda, and I assume its true for hip, but I'll check.

adayton1:
> I know its true for cuda, and I assume its true for hip, but I'll check.

I can't find anything concrete, but it seems to hint at it.

adayton1:
These changes seem unnecessary and are likely to degrade performance.

PR #281, opened 2024-09-11
Tabale reproducer

liu15:
To reproduce crash on rzadams:

mkdir build_rocm
cd build_rocm
cmake -DCHAI_ENABLE_REPRODUCERS=1 -C ../configs/lc/toss_4_x86_64_ib_cray/amdclang.cmake ..
flux alloc -N 1 -n 1 -g1 make -j 40
flux alloc -N 1 -n 1 -g1 ./bin/managed_ptr_multiple_inheritance_reproducer.exe

- Note that the "this" pointer in YofXfromRTTable1D::RootFromBaseX differs from the "this" pointer in other YofXfromRTTable1D methods.
- Interestingly , the crash goes away if GetNumStrings() is removed

The rzansel case does not crash and has consistent pointer addresses in YofXfromRTTable1D.  This can be reproduced with

mkdir build_cuda
cd build_cuda
cmake  -DCHAI_ENABLE_REPRODUCERS=1 -C ../configs/lc/blueos_3_ppc64le_ib_p9/nvcc_clang.cmake ..
lalloc 1 make -j 40
lalloc 1 ./bin/managed_ptr_multiple_inheritance_reproducer.exe


reproducers/managed_ptr_multiple_inheritance_reproducer.cpp
```diff
@@ -0,0 +1,602 @@
...
+      CHAI_HOST_DEVICE virtual ~LinearTable1D() ;
+
+      CHAI_HOST_DEVICE virtual real8 innerEvaluate(real8 x) const override;
+
+   private:
+      CHAI_HOST_DEVICE LinearTable1D() = delete;
+      CHAI_HOST_DEVICE LinearTable1D(const LinearTable1D &other) = delete;
+      CHAI_HOST_DEVICE LinearTable1D &operator=(const LinearTable1D &other) = delete;
+
+      real8 *m_x0 ;
```
adayton1, line 165: Would it be hard to remove some of these data members?

liu15: When I tried removing data members (not this one, but most of the others), I was not able to reproduce the error (but it gets confusing whether I messed something up in the removal or not).

adayton1 (approved):
LGTM

configs/lc/blueos_3_ppc64le_ib_p9/nvcc_clang.cmake
```diff
@@ -0,0 +1,27 @@
...
+set(BLT_CMAKE_IMPLICIT_LINK_DIRECTORIES_EXCLUDE "/usr/tce/packages/gcc/gcc-4.9.3/lib64/gcc/powerpc64le-unknown-linux-gnu/4.9.3;/usr/tce/packages/gcc/gcc-4.9.3/lib64" CACHE STRING "")
+
+set(ENABLE_CUDA ON CACHE BOOL "Enable CUDA")
+set(CUDA_TOOLKIT_ROOT_DIR "/usr/tce/packages/cuda/cuda-11.8.0" CACHE PATH "Path to CUDA")
+set(CMAKE_CUDA_COMPILER "${CUDA_TOOLKIT_ROOT_DIR}/bin/nvcc" CACHE PATH "")
+set(CMAKE_CUDA_HOST_COMPILER "${CMAKE_CXX_COMPILER}" CACHE PATH "")
+set(CMAKE_CUDA_ARCHITECTURES "70" CACHE STRING "")
+set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -Xcompiler=--gcc-toolchain=${GCC_HOME}" CACHE STRING "")
+
+set(ENABLE_REPRODUCERS ON CACHE BOOL "")
```
liu15, line 25: These were set specifically for testing the reproducer.  Should I back these out?  Do you want me to also commit the CARE TOSS4 config?

adayton1: Yeah, I would back out the ENABLE_REPRODUCERS line in these configs.

CMakeLists.txt
```diff
@@ -71,8 +71,6 @@ cmake_dependent_option(CHAI_ENABLE_DOCS "Build CHAI docs" On
                        "ENABLE_DOCS" Off)
 cmake_dependent_option( CHAI_ENABLE_GMOCK "Build CHAI with gmock" On
                         "ENABLE_GMOCK" Off )
-cmake_dependent_option(CHAI_ENABLE_REPRODUCERS "Build CHAI reproducers" On
```
adayton1, line 74: ENABLE_REPRODUCERS is not a variable defined anywhere (unlike ENABLE_TESTS for example, which is a BLT variable). Even if it was, I don't think it makes sense for it to be a dependent variable. It is already an option in cmake/SetupChaiOptions.cmake.

dtaller:
> To reproduce crash on rzadams:
> 
> mkdir build_rocm cd build_rocm cmake -DCHAI_ENABLE_REPRODUCERS=1 -C ../configs/lc/toss_4_x86_64_ib_cray/amdclang.cmake .. flux alloc -N 1 -n 1 -g1 make -j 40 flux alloc -N 1 -n 1 -g1 ./bin/managed_ptr_multiple_inheritance_reproducer.exe
> 
>     * Note that the "this" pointer in YofXfromRTTable1D::RootFromBaseX differs from the "this" pointer in other YofXfromRTTable1D methods.
> 
>     * Interestingly , the crash goes away if GetNumStrings() is removed
> 
> 
> The rzansel case does not crash and has consistent pointer addresses in YofXfromRTTable1D. This can be reproduced with
> 
> mkdir build_cuda cd build_cuda cmake -DCHAI_ENABLE_REPRODUCERS=1 -C ../configs/lc/blueos_3_ppc64le_ib_p9/nvcc_clang.cmake .. lalloc 1 make -j 40 lalloc 1 ./bin/managed_ptr_multiple_inheritance_reproducer.exe

@liu15 . I would put somthing like this (explanation of the reproducer and what systems it fails on and how) as a comment at the top of the reproducer or in a README file somewhere. Those details would be useful to figuring out what it wrong, so it would be nice to have it documented there

dtaller (approved):
Besides adding a readme or big comment explaining the reproducer (as I suggested in my last comment), this looks good.

PR #284, opened 2024-10-29
Remove deprecated getPointer method

adayton1:
Remove deprecated getPointer method. Use data method instead.


dtaller (approved):
Looks good to me

liu15 (approved):

PR #289, opened 2024-11-19
[WIP] set reuse to false in spack environment

adrienbernede:
Testing whether this prevent Spack from using intel for the dependencies of a root spec asking for gcc.


adrienbernede:
closed in favor of #290

PR #295, opened 2025-01-23
Remove incr and decr methods on ManagedArray

adayton1:
These are rarely (never as far as I can tell) used and can easily be replaced using pick and set. Also, incr and decr aren't general for all types.


liu15 (approved):

robinson96 (approved):
LGTM

PR #298, opened 2025-02-03
ManagedSharedPtr

mdavis36:
# Summary

- This pull request introduces `chai::expt::ManagedSharedPtr`, a smart pointer enabling implicit migration of polymorphic types between host and device memory spaces. 
- With `ManagedSharedPtr`, you can safely copy a base shared pointer and invoke derived class functionality using the device-side v-table. 
- The implementation ensures that implicit copies transfer the correct derived object data between host and device.
- Reference counting on the host manages GPU memory, automatically releasing device memory when the object is destroyed.

This has been placed under the `expt` names space per @adayton1's request since much of the implementation is a duplicate of the underlying logic for `chai::ManagedArray`.

This work is necessary for the Spheral GPU port.


src/chai/ManagedArray.inl
```diff
@@ -249,12 +249,14 @@ CHAI_HOST void ManagedArray<T>::reallocate(size_t elems)
       // trigger a moveInnerImpl, which expects inner values to be initialized.
       if (initInner(old_size/sizeof(T))) {
         // if we are active on the  GPU, we need to send any newly initialized inner members to the device
+#if defined(CHAI_ENABLE_CUDA) || defined(CHAI_ENABLE_HIP)
```
adayton1, line 252: `|| defined(CHAI_ENABLE_GPU_SIMULATION_MODE)`

src/chai/ManagedArray.inl
```diff
@@ -302,6 +306,9 @@ CHAI_HOST void ManagedArray<T>::reset()
 template<typename T>
 CHAI_INLINE
 CHAI_HOST_DEVICE size_t ManagedArray<T>::size() const {
+#if !defined(CHAI_DEVICE_COMPILE)
```
adayton1, line 309: I'm concerned about this change. Do you have a reproducer to show why this is necessary?

tests/integration/managed_array_tests.cpp
```diff
@@ -593,7 +591,48 @@ TEST(ManagedArray, ReallocateCPU)
...
+
+  forall(sequential(), 0, 10, [=](int i) {
+      array[i] = i;
+      ASSERT_EQ(&array[i], &array_copy[i]);
+  });
+
+  array.reallocate(20);
+  
+  // This will be incorrect, a call to move, data or copy needs to 
+  // be exectued in order to update the internal active pointer of
```
adayton1, line 609: Spelling: exectued

tests/integration/managed_array_tests.cpp
```diff
@@ -593,7 +591,48 @@ TEST(ManagedArray, ReallocateCPU)
   assert_empty_map(true);
 }
 
+TEST(ManagedArray, ReallocateCopyCPU)
```
adayton1, line 594: Can you provide more context about what this test case is for? I think in general after a reallocate of the original array, any copies made prior to the reallocate are invalidated. The copies might get back in sync after a move is triggered (since they share the same pointer record) in the implementation in ManagedArray.inl, but I don't think this is the case in the thin version where a pointer record is not used.

tests/integration/managed_shared_ptr_tests.cpp
```diff
@@ -0,0 +1,360 @@
...
+#include "chai/managed_ptr.hpp"
+#include "chai/ManagedSharedPtr.hpp"
+
+#include "../src/util/forall.hpp"
+
+// Standard library headers
+#include <cstdlib>
+
+#if defined(CHAI_ENABLE_CUDA)
+inline void gpuErrorCheck(cudaError_t code, const char *file, int line, bool abort=true)
```
adayton1, line 33: Aren't these already defined somewhere? If not, we should add a test utils header for these.

src/chai/ManagedArray.inl
mdavis36: This is probably not necessary for ManagedSharedPtr, this was necessary for the implementation of a ManagedVector like object but it can be necessary for a standar ManagedArray if you aren't careful.

```c++
using chai::ManageArray<double> = MA;
MA a1(10);
MA a2 = a1;

a1.reallocate(20);
assert(a1.data() == a2.data()) // True, we are looking at the same pointer record
assert(a1.size() == a2.size()); //This fails because the internal m_size member isn't updated for a2.
```

The 'size' needs to be held by the MA in order to be able to copy the metadata of the object to the device. However relying on the internal member can cause a disconnect between the allocation copies of a MA are trying to represent.

tests/integration/managed_array_tests.cpp
mdavis36: There is no functionality to "invalidate" a copy of a ManagedArray nor query if it has been invalidated. I strongly believe any copy of a ManagedArray should maintain it's relationship to the underlying allocations it is pointing to, this upholds proper reference semantics copying and is why pointer records exist.

This tests the behavior changed in the `size` member function of a MA seen above. Perhaps this points to a disconnect between the expected behavior of a MA and a gap in the interface. Chai needs to either be explicit about invalidating copies upon reallocation or maintain correct relationships between.

src/chai/ManagedSharedPtr.hpp
```diff
@@ -0,0 +1,331 @@
...
+  template<typename Yp, typename = Compatible<Yp>>
+  CHAI_HOST_DEVICE
+  ManagedSharedPtr(ManagedSharedPtr<Yp> const& rhs) noexcept
+    : m_record_count(rhs.m_record_count)
+    , m_active_pointer(rhs.m_active_pointer)
+    , m_resource_manager(rhs.m_resource_manager)
+  {
+#if !defined(CHAI_DEVICE_COMPILE)
+    if (m_active_pointer) move(ArrayManager::getInstance()->getExecutionSpace());
+    //if (m_active_pointer) move(m_resource_manager->getExecutionSpace());
```
adayton1, line 103: Please remove or document the reason for commented out code.

src/chai/ManagedSharedPtr.hpp
```diff
@@ -0,0 +1,331 @@
...
+  template<typename Tp1>
+  friend class ManagedSharedPtr;
+
+  msp_record_count m_record_count;
+  mutable element_type* m_active_pointer = nullptr;
+
+  mutable SharedPtrManager* m_resource_manager = nullptr;
+
+  template <bool B = is_CHAICopyable<Tp>::value,
+  //template <bool B = std::is_base_of<CHAICopyable, Tp>::value,
```
adayton1, line 214: Please remove.

src/chai/ManagedSharedPtr.hpp
```diff
@@ -0,0 +1,331 @@
...
+            typename std::enable_if<B, int>::type = 0>
+  CHAI_HOST
+  void
+  moveInnerImpl() const 
+  {
+    m_record_count.moveInnerImpl();
+  }
+
+  template <bool B = is_CHAICopyable<Tp>::value,
+  //template <bool B = std::is_base_of<CHAICopyable, Tp>::value,
```
adayton1, line 224: Please remove?

src/chai/SharedPointerRecord.hpp
```diff
@@ -0,0 +1,60 @@
...
+  void* m_pointers[NUM_EXECUTION_SPACES];
+  bool m_touched[NUM_EXECUTION_SPACES];
+  bool m_owned[NUM_EXECUTION_SPACES];
+
+  ExecutionSpace m_last_space;
+  //UserCallback m_user_callback;
+
+  int m_allocators[NUM_EXECUTION_SPACES];
+
+  //msp_pointer_record(void* host_p = nullptr, void* device_p = nullptr) :
```
adayton1, line 41: Remove?

src/chai/SharedPointerRecord.hpp
```diff
@@ -0,0 +1,60 @@
...
+  msp_pointer_record() :
+    m_last_space(CPU) { 
+    for (int space = 0; space < NUM_EXECUTION_SPACES; ++space ) {
+      m_pointers[space] = nullptr;
+      m_touched[space] = false;
+      m_owned[space] = true;
+      m_allocators[space] = 0;
```
adayton1, line 48: I'm wondering if we should default to -1 since I believe 0 is a valid Umpire allocator ID.

src/chai/SharedPtrCounter.hpp
```diff
@@ -0,0 +1,232 @@
...
+  {}
+
+  virtual void m_dispose() noexcept { delete (Ptr)m_record->m_pointers[chai::CPU]; }// TODO : Other Exec spaces...
+  virtual void m_destroy() noexcept { delete this; }
+
+  virtual void moveInnerImpl() const {
+    using T = std::remove_pointer_t<Ptr>;
+    Ptr host_ptr = (Ptr) m_record->m_pointers[CPU]; 
+    // trigger the copy constructor
+    std::cout << "Trigger Inner Copy Ctor @ " << host_ptr << std::endl;
```
adayton1, line 67: Please remove output text.

src/chai/SharedPointerRecord.hpp
mdavis36: This follows the MA pointer record design. I'll add an issue to change this in a follow up PR.

src/chai/SharedPointerRecord.hpp
```diff
@@ -0,0 +1,60 @@
...
+  // Using NUM_EXECUTION_SPACES for the time being, this will help with logical
+  // control since ExecutionSpaces are already defined.
+  // Only CPU and GPU spaces will be used.
+  // If other spaces are enabled they will not be used by ManagedSharedPtr.
+  void* m_pointers[NUM_EXECUTION_SPACES];
+  bool m_touched[NUM_EXECUTION_SPACES];
+  bool m_owned[NUM_EXECUTION_SPACES];
+
+  ExecutionSpace m_last_space;
+  //UserCallback m_user_callback;
```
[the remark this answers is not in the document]
mdavis36: Will add a TODO here and document with an issue.

src/chai/SharedPtrCounter.hpp
```diff
@@ -0,0 +1,232 @@
...
+
+  virtual void m_destroy() noexcept { this->~msp_counted_deleter(); }
+
+  virtual void moveInnerImpl() const {
+
+    using T_non_const = std::remove_const_t<std::remove_pointer_t<Ptr>>;
+    T_non_const* host_ptr = const_cast<T_non_const*>((Ptr)m_impl.m_record->m_pointers[CPU]); 
+
+    // trigger the copy constructor
+    std::cout << "Trigger Inner Copy Ctor @ " << host_ptr << std::endl;
```
adayton1, line 142: Please remove output.

src/chai/SharedPtrCounter.hpp
```diff
@@ -0,0 +1,232 @@
...
+    T_non_const inner = T_non_const(*host_ptr);
+
+    // ensure the inner type gets the state of the result of the copy
+    //err_func(host_ptr);
```
adayton1, line 146: Please remove commented out code.

src/chai/ChaiManager.hpp
```diff
@@ -0,0 +1,28 @@
+//////////////////////////////////////////////////////////////////////////////
```
adayton1, line 1: I'm not sure I see the use for having this as a separate header. I'd prefer this change be reverted.

src/chai/SharedPtrManager.inl
```diff
@@ -0,0 +1,130 @@
...
+#include "chai/config.hpp"
+
+#include "chai/SharedPtrManager.hpp"
+#include "chai/ChaiMacros.hpp"
+
+#include <iostream>
+
+#include "umpire/ResourceManager.hpp"
+
+//#include <cuda_runtime_api.h>
```
adayton1, line 19: Remove?

src/chai/SharedPtrManager.hpp
```diff
@@ -0,0 +1,442 @@
...
+   void set(T* dst_ptr, size_t index, const T& val);
+#endif
+
+  /*!
+   * \brief Get the size of the given pointer.
+   *
+   * \param pointer Pointer to find the size of.
+   * \return Size of pointer.
+   */
+  //CHAISHAREDDLL_API size_t getSize(void* pointer);
```
adayton1, line 165: I would prefer that unimplemented functions or ones we aren't sure are applicable be removed and we just add an issue to track them.

mdavis36: I will go through and review / remove all commented code added in this PR

src/chai/SharedPtrManager.hpp
```diff
@@ -0,0 +1,442 @@
...
+#include "chai/SharedPointerRecord.hpp"
+#include "chai/ChaiManager.hpp"
+
+namespace chai
+{
+namespace expt
+{
+
+/*!
+ * \brief Singleton that manages caching and movement of ManagedArray objects.
```
adayton1, line 19: ... of ManagedSharedPtr objects

src/chai/SharedPtrManager.cpp
```diff
@@ -0,0 +1,626 @@
...
+       CHAI_LOG(Debug, "De-registering " << pointer);
+       m_pointer_map.erase(pointer);
+    }
+  }
+  if (record != &s_null_record) {
+     delete record;
+  }
+}
+
+//void * SharedPtrManager::frontOfAllocation(void * pointer) {
```
adayton1, line 156: Remove and document in an issue. This method might not make sense since with dynamic_casting you may not always get the front of an allocation.

tests/integration/managed_shared_ptr_tests.cpp
```diff
@@ -0,0 +1,360 @@
...
+class NV
+{
+public:
+  unsigned long long content_NV;
+  CHAI_HOST_DEVICE NV(void) : content_NV(0xFFFFFFFFFFFFFFFFull) { printf("++ NV has been constructed\n"); }
+  CHAI_HOST_DEVICE ~NV(void) { printf("-- NV has been destructed\n"); }
+  CHAI_HOST_DEVICE void function(void) const { printf("%llX\n", content_NV); }
+};
+
+GPU_TEST(managed_shared_ptr, shared_ptr_absmem)
```
adayton1, line 170: Since these tests are a little more complicated and the variable naming is not clear to me, could you provide some brief documentation for each test case?

src/chai/ManagedSharedPtr.hpp
```diff
@@ -0,0 +1,331 @@
...
+     ExecutionSpace prev_space = m_record_count.m_get_record()->m_last_space;
+     if (prev_space != GPU && space == GPU) {
+        /// Move nested ManagedArrays first, so they are working with a valid m_active_pointer for the host,
+        // and so the meta data associated with them are updated before we move the other array down.
+       moveInnerImpl();
+     }
+     auto old_pointer = m_active_pointer;
+     m_active_pointer = static_cast<Tp*>(m_resource_manager->move(
+           (void *)m_active_pointer, m_record_count.m_get_record(), space, is_CHAIPoly<Tp>::value));
+     if (old_pointer != m_active_pointer) {
```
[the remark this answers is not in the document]
mdavis36: These comments follow the comments in ManagedArray for explaining moveInnerImpl behavior. I have adjusted them for ManagedSharedPtr context.

src/chai/SharedPtrManager.cpp
```diff
@@ -0,0 +1,626 @@
...
+    new_record->m_owned[space] = true;
+    new_record->m_touched[space] = false;
+  }
+
+  new_record->m_touched[last_space] = true;
+
+  void* dst_pointer = new_record->m_pointers[last_space];
+  void* src_pointer = record->m_pointers[last_space];
+
+  chai::expt::copy(dst_pointer, src_pointer, m_resource_manager, last_space, last_space);
```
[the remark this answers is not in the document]
mdavis36: Only if it is polymorphic. I've added a flag to forward the option.

tests/integration/managed_array_tests.cpp
```diff
@@ -604,12 +603,52 @@ GPU_TEST(ManagedArray, ReallocateGPU)
...
-      ASSERT_EQ(array[i], i);
+      device_assert(array[i] == i);
+    }
+  });
+
+  array.free();
+  assert_empty_map(true);
+}
+
+GPU_TEST(ManagedArray, ReallocateCopyGPU)
```
adayton1, line 619: Could you remove this test as well?

mdavis36:
> Could you add a new CMake configuration variable? Something like CHAI_ENABLE_EXPERIMENTAL? Then in CMakeLists.txt only add the new files if CHAI_ENABLE_EXPERIMENTAL is on.

Done, the option by default disables tests in all of the CI jobs. We probably want to keep testing experimental features so they aren't broken, wheres the best place to define that flag for Chai's CI pipeline?

adayton1 (approved):

src/chai/ManagedArray.hpp
```diff
@@ -373,14 +373,14 @@ class ManagedArray : public CHAICopyable
...
-  // ManagedArrays to nullptr at allocation, since it is extremely easy to
+  // elements with default constructors, since it is extremely easy to
   // trigger a moveInnerImpl, which expects inner values to be initialized.
   template <bool B = std::is_base_of<CHAICopyable, T>::value,
             typename std::enable_if<B, int>::type = 0>
   CHAI_HOST bool initInner(size_t start = 0)
   {
     for (size_t i = start; i < m_size/sizeof(T); ++i) {
-      m_active_base_pointer[i] = nullptr;
+      new (&m_active_base_pointer[i]) T();
```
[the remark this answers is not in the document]
mdavis36: Yes, it allows default construction of more complex nested chai managed types that would be elements of a managed array.

robinson96 (approved):
LGTM

PR #300, opened 2025-02-05
Update copyright to include 2025

adayton1:
Update copyright to include 2025


liu15 (approved):
There has to be a better way to do this!

adayton1:
> There has to be a better way to do this!

There's a script to do it.

dtaller (approved):

PR #308, opened 2025-03-20
const clone

mdavis36:
`clone` should be declared `const` else it can not be called on `const` types.


adayton1 (approved):
Thanks, @mdavis36 !

adayton1:
Do you think this change will warrant a patch release?

adayton1:
Handled by #309 and #310

PR #312, opened 2025-03-21
Update packages

adrienbernede:
- Update RSC (see https://github.com/LLNL/radiuss-spack-configs/pull/126 for changes)
- Optimize submodule usage in GitLab.
- Use radiuss-spack-configs version of ctest-to-junit conversion utility to avoid having to fetch BLT as a submodule.
- Point at specific commit in Spack _before compiler as nodes feature_.
- Update Spack mirror oci options syntax
- Set BLT variable to allow mpi tests to overlap parent allocation (do not run tests on another node).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -35,7 +35,10 @@ variables:
 # at the same location whoever triggers a pipeline.
   CUSTOM_CI_BUILDS_DIR: /usr/workspace/umdev/gitlab-runner
 # Tells Gitlab to recursively update the submodules when cloning the project.
-  GIT_SUBMODULE_STRATEGY: recursive
+  GIT_SUBMODULE_STRATEGY: normal
+  GIT_SUBMODULE_DEPTH: 1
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -35,7 +35,10 @@ variables:
 # at the same location whoever triggers a pipeline.
   CUSTOM_CI_BUILDS_DIR: /usr/workspace/umdev/gitlab-runner
 # Tells Gitlab to recursively update the submodules when cloning the project.
```
adayton1, line 37: Update the comment, too. I guess we don't need submodules of submodules because we are using spack to build the dependencies?

adayton1 (approved):

.gitlab-ci.yml
adrienbernede: Exactly, incidentally, it saves time and traffic through the firewall.

adayton1: Thanks!

liu15 (approved):

PR #332, opened 2025-09-15
v2025.09.1 RC

adayton1:
* Fixes broken experimental build
* Turns off `CHAI_ENABLE_EXPERIMENTAL` by default


src/chai/SharedPtrManager.cpp
```diff
@@ -329,9 +329,9 @@ void SharedPtrManager::free(msp_pointer_record* pointer_record, ExecutionSpace s
 #endif
 #if defined(CHAI_ENABLE_PINNED)
           if (space_ptr == pointer_record->m_pointers[PINNED]) {
-            callback(pointer_record,
-                     ACTION_FREE,
-                     ExecutionSpace(PINNED));
+            //callback(pointer_record,
```
dtaller, line 332: Should you add an explanation of why this is commented out, and when it could be re-enabled?

PR #346, opened 2025-12-22
v2025.12.0 RC -- DO NOT MERGE

adayton1:
This PR is for reviewing purposes only -- to see the changes in the v2025.12.0 RC branch that have not yet been merged into develop. Please review this PR but approve https://github.com/llnl/CHAI/pull/347.


scripts/make_release_tarball.sh
```diff
@@ -7,7 +7,7 @@
 ##############################################################################
 
 TAR_CMD=gtar
-VERSION=2025.09.1
+VERSION=2025.12.0
```
rhornung67: @adayton1 you don't actually need to set this explicitly. There is a GIt command for getting the latest release tag.  Here' what we do in the RAJA version of this script: https://github.com/llnl/RAJA/blob/develop/scripts/make_release_tarball.sh#L11

No need to change this now, just making you aware.

adayton1: I like that. Switching to it now so I don't have to keep changing this script every release.

PR #349, opened 2026-01-09
Add experimental new design for context handling

adayton1:
* All these new classes are experimental.
* Adds Context as a experimental replacement for ExecutionSpace. Unlike ExecutionSpace, it does not mix execution spaces and memory types. It just contains execution spaces (HOST and DEVICE).
* Adds ContextManager to control the current context and synchronization of contexts. This extracts a logical subset of functionality from ArrayManager.
* Adds ContextGuard to set the current context for the remainder of the current scope (similar idea to std::lock_guard)

Future work:
* Make a new ManagedArray class that takes an MemoryManager template argument. All the data coherence is managed by the MemoryManager, so that the new ManagedArray class really just enforces an interface. This will allow multiple different types of memory to be used side-by-side in CHAI applications and really simplify the configuration variables. It will also allow teams to implement their own memory managers if different behavior is desired.
* Separate out dual memory space logic into a DualMemoryManager class
* Separate out single memory space logic into a UnifiedMemoryManager class


tests/expt/ContextGuardTests.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include "gtest/gtest.h"
+
+// Test that ContextGuard updates the current context in scope
+// and restores the previous context on destruction.
+TEST(ContextGuard, ContextGuard) {
+  ::chai::expt::ContextManager& contextManager = ::chai::expt::ContextManager::getInstance();
+  ::chai::expt::Context context = contextManager.getContext();
+
+  {
+    ::chai::expt::Context tempContext = ::chai::expt::Context::HOST;
```
dtaller, line 18: Are there any tests that explicitly use a GPU context or run on the GPU? If not it might be worth adding a GPU test.

dtaller (approved):

liu15 (approved):

tests/expt/ContextGuardTests.cpp
adayton1: Done.

PR #350, opened 2026-01-21
Add experimental ManagedArrayPointer

adayton1:
ManagedArrayPointer is the experimental replacement for ManagedArray. All the logic for handling various memory types is delegated to a ManagerType template parameter.

This class provides a uniform interface for working with different types of memory across multiple backends. It has pointer semantics, meaning that copies are shallow, which allows this object to be passed by value to a CUDA or HIP kernel. When copy constructed, it queries the array manager to update the cached size and pointer from the array manager. The array must be explicitly freed from only one of the shallow copies.

### DIFF
--- a/docs/sphinx/expt/design.rst
+++ b/docs/sphinx/expt/design.rst
@@ -105,3 +105,31 @@ which allows this object to be passed by value to a CUDA or HIP kernel. When cop
 constructed, it queries the array manager to update the cached size and pointer
 from the array manager. The array must be explicitly freed from only one of the
 shallow copies.
+
+.. code-block:: cpp
+
+  #include "chai/expt/ContextRAJAPlugin.hpp"
+  #include "chai/expt/ManagedArrayPointer.hpp"
+  #include "RAJA/RAJA.hpp"
+
+  static ::RAJA::util::PluginRegistry::add<chai::expt::ContextRAJAPlugin> P(
+    "CHAIContextPlugin",
+    "Plugin that integrates CHAI context management with RAJA.");
+
+  // It's recommended to use an alias so that it is easy to swap out the array manager.
+  template <typename T>
+  using ManagedArrayPointer = ::chai::expt::ManagedArrayPointer<T, UnifiedArrayManager>;
+
+  const std::size_t N = 1000000;
+  ManagedArrayPointer<int> a;
+  a.resize(N);
+
+  ::RAJA::forall<::RAJA::seq_exec>(::RAJA::TypedRangeSegment<int>(0, N), [=] (int i) {
+    a[i] = i;  // Use CHAI data structures in the HOST context...
+  });
+
+  constexpr int BLOCK_SIZE = 256;
+
+  ::RAJA::forall<::RAJA::cuda_exec_async<BLOCK_SIZE>>(::RAJA::TypedRangeSegment<int>(0, N), [=] __device__ (int i) {
+    a[i] -= 1;  // Use CHAI data structures in the DEVICE context...
+  });
--- a/docs/sphinx/expt/design.rst
+++ b/docs/sphinx/expt/design.rst
@@ -102,6 +102,6 @@ ManagedArrayPointer
 This class provides a uniform interface for working with different types of memory
 across multiple backends. It has pointer semantics, meaning that copies are shallow,
 which allows this object to be passed by value to a CUDA or HIP kernel. When copy
-constructed, it queries the array manager to update the cached size and pointer.
+constructed, it queries the array manager to update the cached size and pointer
 from the array manager. The array must be explicitly freed from only one of the
 shallow copies.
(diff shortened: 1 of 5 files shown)


src/chai/expt/ManagedArrayPointer.hpp
```diff
@@ -0,0 +1,251 @@
...
+       * @details On host builds, if a manager is present and provides a non-null data pointer,
+       * refreshes the cached pointer `m_data` from `m_manager->data()`. On device builds
+       * (CHAI_DEVICE_COMPILE), returns the cached pointer without querying the manager.
+       */
+      CHAI_HOST_DEVICE ElementType* data() const
+      {
+#if !defined(CHAI_DEVICE_COMPILE)
+        if (m_manager)
+        {
+          if (ElementType* data = static_cast<ElementType*>(m_manager->data()); data)
```
dtaller, line 173: Did you intend to write data again at the end of the if after the semicolon? Add comment

liu15 (approved):

src/chai/expt/ManagedArrayPointer.hpp
dtaller: On I understand you are checking if it is nullptr effectively?

adayton1: Yes, this is c++17 syntax.

src/chai/expt/ManagedArrayPointer.hpp
```diff
@@ -0,0 +1,251 @@
...
+       * @param i Element index.
+       *
+       * @return Reference to element i in the cached data pointer.
+       *
+       * @note No bounds checking is performed.
+       * @note Uses the cached pointer `m_data` and does not call update().
+       */
+      CHAI_HOST_DEVICE ElementType& operator[](std::size_t i) const
+      {
+        return m_data[i];
```
dtaller, line 217: Do you think there should be a debug mode that checks m_data is not null and i is less than size? Not really required, but maybe could be useful debugging

dtaller (approved):

src/chai/expt/ManagedArrayPointer.hpp
adayton1: I'm not sure. On the host, asan should catch out of bounds accesses. HIP and CUDA are supporting asan or asan like functionality.

dtaller: If asan will detect this on both the CPU and GPU then this should be fine then, don't worry about it then

src/chai/expt/ManagedArrayPointer.hpp
```diff
@@ -0,0 +1,251 @@
...
+       *
+       * @details On host builds, synchronizes the cached size from the manager (if present).
+       * On device builds (CHAI_DEVICE_COMPILE), returns the last cached size.
+       */
+      CHAI_HOST_DEVICE std::size_t size() const
+      {
+#if !defined(CHAI_DEVICE_COMPILE)
+        if (m_manager)
+        {
+          m_size = m_manager->size()/sizeof(ElementType);
```
dtaller: Size is being used in two different senses, num bytes vs num elements. Do you think that m_size should be renamed to m_num_elems, or that m_manager->size() should be renamed to m_manager->num_bytes(), or both? To clarify thing a bit?

adayton1: I will rename the manager method to size_bytes (std::span sets a precedence for this name).

dtaller: Thanks!

adayton1: Done.

robinson96 (approved):
LGTM

docs/sphinx/expt/design.rst
```diff
@@ -94,3 +94,42 @@ may be handled by CHAI.
...
+
+-------------------
+ManagedArrayPointer
+-------------------
+
+This class provides a uniform interface for working with different types of memory
+across multiple backends. It has pointer semantics, meaning that copies are shallow,
+which allows this object to be passed by value to a CUDA or HIP kernel. When copy
+constructed, it queries the array manager to update the cached size and pointer
+from the array manager. The array must be explicitly freed from only one of the
```
rhornung67, line 106: Is there a way for a user to check whether the data referenced by a ManagedArrayPointer has been freed; e.g., by another pointer object pointing to the same data?  If not, it may be worth considering. I can see how that would be useful.

rhornung67 (approved):
Looks good to me. One comment to consider.

Also, you may want to rework the copyright statements in the files like I did in RAJA. Then, updating the copyright year only requires touching 2 files instead of all of them, which kinda pollutes the commit history of each file.

Here's the change: https://github.com/llnl/RAJA/pull/1976

docs/sphinx/expt/design.rst
adayton1: Not without a lot of extra tracking. My plan was to add another class called ManagedArraySharedPointer that would act like std::shared_ptr and avoid the need to explicitly free.

rhornung67: Yeah. I figured that would be the case.  You solution sounds like it should do the trick.

PR #351, opened 2026-01-23
Add experimental HostArrayManager

adayton1:
This class is the experimental replacement for the thin host only logic contained in ManagedArray.

This class manages a host array. It is designed for use with ManagedArrayPointer.

HostArrayManager performs value initialization of each array element. That is to say, numeric types will be initialized to zero and nontrivial types will be default constructed. In the future, this behavior may change to default initialization for performance reasons, such that numeric types will be left in an indeterminate state and nontrivial types will be default constructed.

HostArrayManager does not rely on ContextManager, so it will behave differently than other array managers. The major difference is that when used with ManagedArrayPointer, the ManagedArrayPointer does not need the update method called or to be copy constructed before it can be used on the host. Since it will not respect the current Context, be extra careful to avoid using it on the device.

```
#include "chai/expt/HostArrayManager.hpp"
#include "chai/expt/ManagedArrayPointer.hpp"

// It's recommended to use an alias so that it is easy to swap out the array manager.
template <typename T>
using HostArrayManager = ::chai::expt::HostArrayManager<T>;

template <typename T>
using ManagedArrayPointer = ::chai::expt::ManagedArrayPointer<T, HostArrayManager<T>>;

const std::size_t N = 1000000;
ManagedArrayPointer<int> a{HostArrayManager<int>(N)};

for (std::size_t i = 0; i < N; ++i)
{
  a[i] = i;
}
```

Also note that the implementation of ManagedArrayPointer changed slightly since it was determined that the managers need to be templated on ElementType (for correct initialization of the underlying arrays).

Another slight change is that one of the ManagedArrayPointer constructors changed from taking a pointer to taking the Manager by value. If used with temporaries, no deep copies will be performed.

Also adds a factory function called `make` to ManagedArrayPointer.


liu15 (approved):

src/chai/expt/HostArrayManager.hpp
```diff
@@ -0,0 +1,130 @@
...
+      {
+        return m_storage.size();
+      }
+
+      /*!
+       * \brief Returns a pointer to the underlying contiguous storage.
+       *
+       * \return Pointer to the first element, or nullptr if the storage is empty.
+       */
+      ElementType* data()
```
dtaller, line 117: Do you also want a cdata() method (returning const ElementType*) ? We have that for some types in care, but I am not sure whether you want that here?

dtaller (approved):

src/chai/expt/HostArrayManager.hpp
adayton1: I'm not yet sure how I want to add that, though it is on my radar.

PR #352, opened 2026-01-26
Add experimental ManagedArraySharedPointer

adayton1:
This class is an array interface like ManagedArrayPointer, except that it acts like std::shared_ptr such that underlying memory is cleaned up automatically when the last host copy is destroyed (device copies do not and should not count since memory allocation calls are launched from the host).

### DIFF
--- a/docs/sphinx/expt/design.rst
+++ b/docs/sphinx/expt/design.rst
@@ -134,6 +134,47 @@ shallow copies.
     a[i] -= 1;  // Use CHAI data structures in the DEVICE context...
   });
 
+  a.free();
+
+-------------------------
+ManagedArraySharedPointer
+-------------------------
+
+This class provides a uniform interface for working with different types of memory
+across multiple backends. It has shared pointer semantics, meaning that copies are
+shallow, which allows this object to be passed by value to a CUDA or HIP kernel.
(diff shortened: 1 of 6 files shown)


liu15 (approved):

docs/sphinx/expt/design.rst
```diff
@@ -134,6 +134,47 @@ shallow copies.
...
+ManagedArraySharedPointer
+-------------------------
+
+This class provides a uniform interface for working with different types of memory
+across multiple backends. It has shared pointer semantics, meaning that copies are
+shallow, which allows this object to be passed by value to a CUDA or HIP kernel.
+When copy constructed, it queries the array manager to update the cached size and
+pointer from the array manager. Like with std::shared_ptr, the underlying memory
+will be cleaned up automatically once the last host copy is destroyed.
+
```
rhornung67, line 151: I think it would be helpful to add something about how device instances are not reference counted, etc. similar to what you have in the PR description.

dtaller (approved):

docs/sphinx/expt/design.rst
adayton1: Thanks for the suggestion! Done.

PR #356, opened 2026-02-25
Use project specific variable to specify github status token

adrienbernede:
Fix issue with machine outage report.

The issue identified is that token cannot be accessed correctly when defined at group level anymore.

Instead, we defined the token at project level.


adayton1 (approved):

liu15 (approved):

adrienbernede:
@adayton1 You may ignore the remaining failure status merge: it’s outdated and this type of situation is fixed in https://github.com/llnl/CHAI/pull/355.

PR #357, opened 2026-02-25
Prevent resource manager from being created by empty managed arrays

liu15:
This prevents pre-main cuda calls for global static Managed Arrays.

### DIFF
--- a/src/chai/ManagedArray.inl
+++ b/src/chai/ManagedArray.inl
@@ -25,7 +25,6 @@ CHAI_HOST_DEVICE ManagedArray<T>::ManagedArray():
   m_is_slice(false)
 {
 #if !defined(CHAI_DEVICE_COMPILE)
-  m_resource_manager = ArrayManager::getInstance();
   m_pointer_record = &ArrayManager::s_null_record;
 #endif
 }
@@ -37,6 +36,7 @@ ManagedArray<T>::ManagedArray(
     std::initializer_list<umpire::Allocator> allocators):
   ManagedArray()
 {
+  m_resource_manager = ArrayManager::getInstance();
   m_pointer_record = new PointerRecord();
   int i = 0;
   for (int s = CPU; s < NUM_EXECUTION_SPACES; ++s) {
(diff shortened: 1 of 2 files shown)


liu15:
There is a bug here somewhere.  I am investigating.

liu15:
> There is a bug here somewhere. I am investigating.

NVM, I think I am hitting existing stale data issues from gpu_sim mode.  I don't think I will be able to test this properly until those are fixed.

src/chai/ManagedArray.inl
```diff
@@ -169,6 +169,9 @@ CHAI_HOST void ManagedArray<T>::allocate(
     const UserCallback& cback) 
 {
   if(!m_is_slice) {
+     if (m_resource_manager == nullptr) {
```
adayton1, line 172: I'm not sure I really like these changes where almost any operation now has to check if the ArrayManager exists. The class no longer establishes an invariant.

adayton1: If we really have to, a getter would be much better.

adayton1 (changes requested):

src/chai/ManagedArray.inl
liu15: I can probably remove the calls from pick and set, since those should only be valid for allocated arrays. For the registerTouch and reset functions, perhaps they should just return if there is no resource manager?

adayton1: What about reallocate? I just don't like that we have to decide on a per function basis what the state of the object is supposed to be.

adayton1: Would it be possible to push this up the chain to ArrayManager? Have ArrayManager not make any pre-main calls until it has to?

liu15: reallocate only need the resource manager if it is already allocated.  Basically, the resource manage is present if the managed array has been allocated at any point.

liu15: > Would it be possible to push this up the chain to ArrayManager? Have ArrayManager not make any pre-main calls until it has to?

I guess that could be possible.  Would you suggest making ArrayManager::m_resource_manager a getter?

liu15: It it was a getter, would it be better in ManagedArray or ArrayManager?

adayton1 (approved):

robinson96 (approved):
LGTM

src/chai/ManagedArray.inl
adayton1: We decided no getter is needed. If the ManagedArray is allocated, then it will actually point to the ArrayManager. Some preconditions were added to some functions to indicate they should only be called on an allocated ManagedArray.

PR #380, opened 2026-07-24
Fix overload ambiguity errors

adayton1:
`chai::ManagedArray` had two constructors that sometimes led to compilation errors due to ambiguous overload resolution.

```
CHAI_HOST_DEVICE ManagedArray(std::nullptr_t>) : ManagedArray() {}

CHAI_HOST_DEVICE explicit ManagedArray(
      size_t elems,
      ExecutionSpace space = get_default_space());
```

Compilation would fail when constructing a `ManagedArray` from the integer literal `0` because the compiler can't decide which overload to pick. Additionally, nvcc erroneously reports ambiguity when constructing from an int in certain cases.

Using a constraint resolves the ambiguity in both of those situations. The nullptr constructor will now only accept `nullptr` or a variable of type `std::nullptr_t`. The size_t constructor will get all the rest, including `0`.

### DIFF
--- a/.gitlab/jobs/corona.yml
+++ b/.gitlab/jobs/corona.yml
@@ -17,6 +17,12 @@
 # project. We keep ${PROJECT_<MACHINE>_VARIANTS} and ${PROJECT_<MACHINE>_DEPS}
 # So that the comparison with the original job is easier.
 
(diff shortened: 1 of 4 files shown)


dtaller (approved):

liu15 (approved):

adayton1 (commented):
@lc-hubcast approve

PR #391, opened 2026-09-15
Add a PointerTable unpacker

robinson96:
Add an unpacker for managed_ptr that helps manage an array of pointers in managed_ptr objects.


liu15 (approved):

src/chai/managed_ptr.hpp
```diff
@@ -962,7 +962,175 @@ namespace chai {
...
+            }
+
+            auto* arrayManager = chai::ArrayManager::getInstance();
+            m_cpu_allocator_id = arrayManager->getAllocatorId(CPU);
+            auto cpuAllocator = arrayManager->getAllocator(m_cpu_allocator_id);
+            m_cpu_pointers = static_cast<T**>(cpuAllocator.allocate(m_size * sizeof(T*)));
+
+            using pointer_type = T*;
+            for (size_t index = 0; index < m_size; ++index) {
+               ::new (static_cast<void*>(m_cpu_pointers + index)) pointer_type(managedPointers[index].get(CPU));
```
adayton1: You need to ensure managedPointers is up-to-date on the host instead of accessing it directly. I would suggest something like:

```
chai::managed_ptr<T>* hostManagedPointers = managedPointers.data();
```

and then using `hostManagedPointers[index]` in this loop.

src/chai/managed_ptr.hpp
```diff
@@ -962,7 +962,175 @@ namespace chai {
...
+         PointerTable(const PointerTable&) = delete;
+         PointerTable& operator=(const PointerTable&) = delete;
+
+         CHAI_HOST ~PointerTable()
+         {
+            auto* arrayManager = chai::ArrayManager::getInstance();
+
+#if (defined(CHAI_GPUCC) || defined(CHAI_ENABLE_GPU_SIMULATION_MODE)) && defined(CHAI_ENABLE_MANAGED_PTR_ON_GPU)
+            if (m_gpu_pointers != nullptr) {
+               arrayManager->getAllocator(m_gpu_allocator_id).deallocate(m_gpu_pointers);
```
adayton1, line 1087: Since you call placement new, you also need to explicitly call the destructor.

src/chai/managed_ptr.hpp
```diff
@@ -962,7 +962,175 @@ namespace chai {
...
+            }
+#endif
+            if (m_cpu_pointers != nullptr) {
+               arrayManager->getAllocator(m_cpu_allocator_id).deallocate(m_cpu_pointers);
```
adayton1, line 1091: Since you call placement new, you also need to explicitly call the destructor.

src/chai/managed_ptr.hpp
robinson96: The type of managed pointers is a raw array of managed_ptr, so I 'm pretty sure this is correct as is.

src/chai/managed_ptr.hpp
robinson96: the corresponding allocation is L1058, which is an allocator allocate call.

src/chai/managed_ptr.hpp
robinson96: The corresponding allocation is L1048, which is an allocator allocate call.

src/chai/managed_ptr.hpp
```diff
@@ -962,7 +962,175 @@ namespace chai {
...
+         CHAI_HOST_DEVICE PointerTableView(T** cpuPointers, T** gpuPointers)
+            : m_cpu_pointers(cpuPointers),
+              m_gpu_pointers(gpuPointers)
+         {
+         }
+
+         CHAI_HOST_DEVICE T** data() const
+         {
+#if defined(CHAI_DEVICE_COMPILE) && defined(CHAI_ENABLE_MANAGED_PTR_ON_GPU)
+            return m_gpu_pointers;
```
adayton1, line 986: Do we need to worry about this data going stale?

src/chai/managed_ptr.hpp
adayton1: const chai::ManagedArray<chai::managed_ptr<T>>& managedPointers

adayton1: For GPU construction, I think you do have a raw pointer which you get by calling managedPointers.data(GPU).

src/chai/managed_ptr.hpp
adayton1: I'm talking about the call at L1052 that also needs to be addressed.

src/chai/managed_ptr.hpp
adayton1: The call at L1008 or 1019 also needs a matching destructor call.

src/chai/managed_ptr.hpp
robinson96: I mean, not more than any other managed_ptr creation / unpack mechanism? I think the user of this needs to be mindful of what they're doing.

adayton1: I guess what I mean is that you should document any limitations to the new unpacker (mainly the the `unpack_table_pointers` function), including (I believe from what I've seen) that it will not reflect changes made to the inner `managed_ptr<T>`objects after the call to unpack.

src/chai/managed_ptr.hpp
robinson96: Sorry, must have been looking at the GPU construction.

src/chai/managed_ptr.hpp
```diff
@@ -962,7 +962,173 @@ namespace chai {
...
+               return;
+            }
+
+            auto* arrayManager = chai::ArrayManager::getInstance();
+            m_cpu_allocator_id = arrayManager->getAllocatorId(CPU);
+            auto cpuAllocator = arrayManager->getAllocator(m_cpu_allocator_id);
+            m_cpu_pointers = static_cast<T**>(cpuAllocator.allocate(m_size * sizeof(T*)));
+            chai::managed_ptr<T>* hostManagedPointers = managedPointers.data();
+
+            for (size_t index = 0; index < m_size; ++index) {
```
adayton1, line 1079: Maybe use/adapt populate_pointer_table?

adayton1 (approved):

adayton1 (approved):

adayton1 (commented):
@lc-hubcast approve

src/chai/managed_ptr.hpp
adayton1: New documentation looks good.

adayton1 (approved):

adayton1 (commented):
@lc-hubcast approve

adayton1 (commented):
@lc-hubcast approve

